PR #10, opened 2015-07-14
change mageia template to use python-devel

joequant:
Mageia still uses python-devel as the dependency name rather
than python2-devel.  This changes the mageia template to
modify the output of the macro.


mcyprian:
Thank you, I accept this PR. We don't test Mageia template, I suppose your changes are correct.

PR #15, opened 2015-07-19
fix mageia python2/3 generation string

joequant:
mageia still uses python for python2 and python3 for
python3.  This patch fixes the bin and site_arch
string generation.


mcyprian:
Merged, thank you

Issue #16, opened 2015-07-28
cli script is missing from %files section when -p3 option is used


marbu:
When I try to regenerate srpm of kaptan with python3 support like this
(ignoring issue #7  for now):

```
pyp2rpm -n kaptan --srpm -p3
```

pyp2rpm doesn't include kaptan command line tool into `%files` section which
means that generated spec file is invalid:

```
Checking for unpackaged file(s): /usr/lib/rpm/check-files /home/makerpm/rpmbuild/BUILDROOT/python-kaptan-0.5.8-1.fc22.i386
error: Installed (but unpackaged) file(s) found:
   /usr/bin/kaptan
   /usr/bin/python3-kaptan


RPM build errors:
    Installed (but unpackaged) file(s) found:
   /usr/bin/kaptan
   /usr/bin/python3-kaptan
```

Relevant diff of generated specfile follows (compared to previous run without `-p3`):

```
 %files
 %doc README.md LICENSE
-%{_bindir}/kaptan
 %{python2_sitelib}/%{pypi_name}
 %{python2_sitelib}/%{pypi_name}-%{version}-py?.?.egg-info
+%if 0%{?with_python3}
+%files -n python3-%{pypi_name}
+%doc README.md LICENSE
+%{python3_sitelib}/%{pypi_name}
+%{python3_sitelib}/%{pypi_name}-%{version}-py?.?.egg-info
+%endif # with_python3
+
```

Unfortunately since python packaging policy is being redefined because of
python3 as default change, I'm not sure what the expected fix for this would
be.

I'm using `pyp2rpm-1.1.2-1.fc22.noarch`

PR #18, opened 2015-08-10
New fedora template

mcyprian:
Packaging Python guidelines now recommends to build both python2 and python3 versions in the same directory and to use new %py2_build and %py2_install macros. This PR adds new fedora template and moves the old to file fedora_subdirs.spec.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,10 @@
 language: python
 python:
+    - "2.6"
     - "2.7"
     - "3.4"
 env:
+    -TOXENV=py26
     -TOXENV=py27
     -TOXENV=py34
 
--- a/README.md
+++ b/README.md
@@ -60,6 +60,12 @@ To run the unit tests, cd into the checked out directory and run::
 or run::
 
     python setup.py test
+    
+Recommended way to use:
+
+![alt tag](https://mcyprian.fedorapeople.org/pyp2rpm_guide2.gif 
+"Record of pyp2rpm usage")
+
 
 I will gladly accept any pull request or recommendation.
 With complex pull requests, please include unit tests in *pytest*, use *flexmock* if you need mocking.
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ or run::
     
 Recommended way to use:
 
-![alt tag](https://mcyprian.fedorapeople.org/pyp2rpm_guide2.gif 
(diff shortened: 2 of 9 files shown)


rkuska:
I think it is enough to generate documentation with sphinx only once (if available ofc), also it should be moved to `%build` or `%install` phase of the specfile.

**So if sphinx is available -**
generate documentation only once (using sphinx-build or sphinx-build-3 based on base python interpret in specfile)
move it to `%build` or `%install`  phase
create subpackage python-foo-doc which will ship docs generated by sphinx

cc @bkabrda for opinion

running `./mybin.py devassistant -b 3`

pyp2rpm finds following binary files
%{_bindir}/python3-da
%{_bindir}/python3-da-gui
%{_bindir}/python3-devassistant
%{_bindir}/python3-devassistant-gui

these shoud be without `python3-` prefix. 

**How executables should be treated:**

when pyp2rpm is run without additional parameters (both python3 and python2 subpackage is created and python2 is base python), assume we have executable foo:
`foo`, `foo-2`, `foo-2.7` should be packaged under python2 subpackage
`foo-3`, `foo-3.4` should be packaged under python3 subpackage

when pyp2rpm is run with -b 3 (only python3 subpackage is generated and also python3 is base python)
`foo`, `foo-3`, `foo-3.4` should be packaged under python3 subpackage

when pyp2rpm is run with -b 3 -p 2 (both python3 and python2 subpackage is created and python3 is base python)
`foo-2`, `foo-2.7` should be packaged under python2 subpackage
`foo`, `foo-3`, `foo-3.4` should be packaged under python3 subpackage  
_`-X.Y` version can be generated by `%{pythonX-version}` macro_

sYnfo:
> I think it is enough to generate documentation with sphinx only once (if available ofc), also it should be moved to %install phase of the specfile.

Wouldn't %build be more appropriate, given that you're building docs?

rkuska:
> Wouldn't %build be more appropriate, given that you're building docs?

I agree, `%build` seems much more reasonable place for **build**ing docs.

mcyprian:
Sphinx build moved to %build section as you recommended, executable naming updated. I've found out that variable data.scripts was map object instead of list, what was the reason of behavior mentioned in issue #16, so I've fixed it as well.

rkuska:
One more thing, Requires are duplicated between main package and subpackage which carries python version marked as base.

PR #19, opened 2015-08-25
Virtualenv module

mcyprian:
It is possible to get more accurate metadata using this module, it installs package to virtualenv in temporary directory and scans scripts,  and site_pacakges file that appears after instalation. It can also get list of other installed packages - package runtime dependancies. Metadata from virtualenv are appended to metadata from archive and don't overwrite them.

### DIFF
--- a/pyp2rpm/virtualenv.py
+++ b/pyp2rpm/virtualenv.py
@@ -109,7 +109,7 @@ class VirtualEnv(object):
         formated_deps = []
         for dep in deps_list:
             name = self.name_convertor.rpm_name(dep[0])
-            formated_deps.append(['Requires', name])
+            formated_deps.append(['Requires', name.lower()])
         return formated_deps
 
     @property
--- a/pyp2rpm/virtualenv.py
+++ b/pyp2rpm/virtualenv.py
@@ -4,6 +4,7 @@ import locale
 from subprocess import Popen, PIPE, DEVNULL
 
 from pyp2rpm.settings import VENV_INTERPRETER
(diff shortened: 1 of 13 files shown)


pyp2rpm/virtualenv.py
```diff
@@ -0,0 +1,159 @@
...
+        self.dirs_after_install = DirsContent()
+        self.dirs_before_install.fill(self.temp_dir + '/venv/')
+        self.data = {}
+
+    def create_virtualenv(self):
+        '''
+        Creates new virtualenv in temp_dir
+        '''
+        proc = Popen(['virtualenv', '-p', VENV_INTERPRETER, self.temp_dir + '/venv/'],
+                     stdout=DEVNULL, stderr=PIPE)
```
sYnfo: Have you considered using the virtualenv-api package? It seems it would be a good fit for this usecase.

[0] https://pypi.python.org/pypi/virtualenv-api/

mcyprian:
I 've rewritten it using virtualenv-api, code doesn't seem to be much faster but it is definitely a bit nicer. I 've also added tests for this module and made it all Python2.6 compatible.

rkuska:
Trying `pyp2rpm pytest` I end up with:

```
%{_bindir}/py.test
%{_bindir}/py.test-2
%{_bindir}/py.test-%{python2_version}
%{_bindir}/py.test-2.7
%{_bindir}/py.test-2.7-2
%{_bindir}/py.test-2.7-%{python2_version}
```

PR #21, opened 2015-10-01
extract python version from setup / egg

MichaelMraka:
Some packages specifies python 3 in their setup.py / egg and can't be built on python 2.
This patch extracts proper python version from package classifiers.

### DIFF
--- a/pyp2rpm/bin.py
+++ b/pyp2rpm/bin.py
@@ -65,14 +65,6 @@ def main(package, v, d, r, proxy, srpm, p, b, o, t):
     PACKAGE             Provide PyPI name of the package or path to compressed source file."""
     register_file_log_handler('/tmp/pyp2rpm-{0}.log'.format(getpass.getuser()))
 
-    if not p and not b:
-        p = settings.DEFAULT_ADDITIONAL_VERSION
-    if not b:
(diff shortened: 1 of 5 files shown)


hroncok:
Seems useful :+1:

rkuska:
Hi thank you for your contribution! This indeed seems useful, would you also provide unittests with your PR? Thank you!

mcyprian:
When using base python option: **pyp2rpm nose -b 3** the result is: 

``` bash
%files -n python3-%{pypi_name} 
%doc README.txt
%dir %{python3_sitelib}/__pycache__/
%{python3_sitelib}/__pycache__/*
%{python3_sitelib}/htmlplug.py
%{python3_sitelib}/%{pypi_name}-%{version}-py?.?.egg-info

%files -n python3-%{pypi_name} 
%doc README.txt
%dir %{python3_sitelib}/__pycache__/
%{python3_sitelib}/__pycache__/*
%{python3_sitelib}/htmlplug.py
%{python3_sitelib}/%{pypi_name}-%{version}-py?.?.egg-info
```

Extraction of versions should be active only if base and additional versions (options -b -p) are not specified explicitly.

PR #26, opened 2015-12-13
fix processing of tarballs

marbu:
This pull requrest fixes processing of `*.tgz` tarballs (generated specfile had correct structure, but didn't contain actuall content) and introduces support for `*.tar.xz` tarballs (`tarball` module from python stdlib contains xz support since python 3.3).


mcyprian:
Merged, thank you

PR #38, opened 2016-04-11
Epel template

### LINKED ISSUE #25: support for rhel/centos distrubutions [feature proposal]
I would be nice if pyp2rpm could create spec files following packaging policies for rhel/centos 6 and 7 distributions. Even though those packaging policies are based on older fedora ones, I'm not sure if just adding proper template would be enough.

How is building packages for rhel/centos connected to fedora? Here is the use case I have in mind:
author of new python module would like to build packages for centos and fedora in copr. Note that copr can build packages for rhel/centos even though it's a fedora project.

mcyprian:
Prototype of EPEL template, fixes #25:
- doesn't use modern macros (%py2_install %py2_build)
- forbids base version > 2 if template is used
- disables with_pythonX macro for versions > 2

### DIFF
--- a/pyp2rpm/bin.py
+++ b/pyp2rpm/bin.py
@@ -126,10 +126,7 @@ def main(package, v, d, s, r, proxy, srpm, p, b, o, t, venv):
 
         if srpm:
             msg = utils.build_srpm(spec_path, d)
-            if utils.PY3:
-                logger.info(msg.decode('utf-8'))
-            else:
-                logger.info(msg)
+            logger.info(msg)
 
     else:
         logger.debug('Printing specfile to stdout.')
--- a/pyp2rpm/convertor.py
+++ b/pyp2rpm/convertor.py
@@ -61,6 +61,29 @@ class Convertor(object):
                 and not os.path.isdir(self.package):
             self.pypi = False
 
+    def merge_versions(self, data):
+        """Merges python versions specified in command lines options with
+        extracted versions, checks if some of the versions is not > 2 if EPEL6 template
+        will be used.
+        """
+        if self.template == "epel6.spec":
+            # if user requested version greater than 2, writes error message and exits
+            if (any(int(ver[0]) > 2 for ver in self.python_versions
+                    + ([self.base_python_version] if self.base_python_version else [])
+                    + [data.base_python_version])):
+                sys.stderr.write("Invalid version, major number of python version for EPEL6 "
+                                 "spec file must not be greater than 2.\n".format(self.base_python_version))
+                sys.exit(1)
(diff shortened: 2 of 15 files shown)


hroncok:
Python 3 EPEL packages are not considered here?

hroncok:
Can you include some nice integration test in this PR? So I can see it in action without even trying to use it? :D

mcyprian:
I was not able to find complete and clear description of python packaging in EPEL.
[EPEL Packaging wiki page](https://fedoraproject.org/wiki/EPEL:Packaging) says that new macros are not available in EL5 and EL6.
I used an EPEL specfile from package db where Python3 is disabled as an inspiration of this template's design. 

If Python 3 can be used in EPEL packages, can't we simply use [fedora template](https://github.com/fedora-python/pyp2rpm/blob/master/pyp2rpm/templates/fedora.spec) or it's older version [fedora_subidirs.spec](https://github.com/fedora-python/pyp2rpm/blob/master/pyp2rpm/templates/fedora_subdirs.spec) for EPEL packages as well?

Do you think we want something more like spec example on this [wiki page](https://fedoraproject.org/wiki/User:Mstuchli/Python3inEPEL)?

[Here](https://paste.fedoraproject.org/354012/03815151/) is an example of spec file generated running **pyp2rpm coverage -p3 -p35 -s -t epel**. I am not sure if this is what you meant by integration test.

marbu:
I'm quite likely to be missing something here, but don't we need a separate template for each EPEL5, EPEL6 and EPEL7? Maybe EPEL5 and EPEL6 are similar enough to share the template, but if I read [Packaging:Python_Old](https://fedoraproject.org/wiki/Packaging:Python_Old) (linked from User:Mstuchli/Python3inEPEL) right, EPEL7 provides significant new features (especially when it comes to python3 support).

hroncok:
By integration test I mean a test in the testsuite that would test if this works. I can talk to you in person on Wednesday abut his.

mcyprian:
@hroncok: Do you have an idea why some of the tests fails with traceback ImportError: No module named jinja2 despite the fact Jinja2 is on the list of installed deps?

hroncok:
No idea

mcyprian:
Integration tests have been fixed. I tested new templates on small set of packages that are directly buildable when fedora.spec template is used. Many packages were built successfully for epel6 and epel7 from the spec files generated using new templates. Some of the packages failed because of missing dependencies in epel repositories. Examples of generated spec files are in files /test/test_data/python-Jinja2_epel6.spec and /test/test_data/python-Jinja2_epel7.spec. 
Do you have some comments/suggestions regarding design of the templates @hroncok @marbu?

PR #52, opened 2016-05-25
man page for pyp2rpm

rsuste:


hroncok:
Thanks for contributing.

Did you just write it in groff syntax or used some tool to generate it? It would be better to generate it, so it is not out of date soon.

rsuste:
I wrote it in groff. Do you want it remake to Markdown or something other?
Dne 25.5.2016 12:03 napsal uživatel "Miro Hrončok" <notifications@github.com

hroncok:
Not really, that would keep the problem. I was thinking generating it from the --help or something like that.

xsuchy:
@hroncok I advised @RadimS97 to not generate it from --help as he wrote there more information than is available in --help output. E.g. Example section.

hroncok:
And that's a good thing, but I'm afraid that the manpage will get outdated, so I'm thinking if there is a way to keep it up to date automatically somehow, but maybe I'm just overthinking it.

pyp2rpm.1
```diff
@@ -0,0 +1,95 @@
...
+.TP
+.B pyp2rpm ~/rpmbuild/SOURCES/python_package_name.tar.gz
+Pick up package from the local filesystem.
+
+ 
+
+	
+
+.SH LANGUAGE
+PYTHON
```
hroncok, line 88: Is this a common section? I don't see it's purpose.

pyp2rpm.1
```diff
@@ -0,0 +1,95 @@
...
+
+.SH SYNOPSIS
+.B pyp2rpm
+[-h] [-v VERSION] [-d SAVE_DIR] [-r RPM_NAME] [-t TEMPLATE]
+[-o DISTRO] [-b BASE_PYTHON] [-p PYTHON_VERSION] [--srpm] [--proxy PROXY]
+.B PACKAGE
+
+
+.SH ARGUMENTS
+.B PACKAGE
```
hroncok, line 15: Maybe use .TP?

xsuchy:
> but maybe I'm just overthinking it.

:) Most of the projects have man page as separate file. And yes, it is one more thing to think about when you add feature/option that you need to write up documentation. But that is life and it happen when some project grown up.

mcyprian:
@hroncok Do you have some more comments to this? Can I merge this PR now?

hroncok:
I've added 2 inline comments, no response, feel free to consider them irrelevant and marge anyway, or change it after the merge.

PR #55, opened 2016-06-22
Description processing improvement

mcyprian:
Add regular expressions to remove PyPI's download and version tags from description [(python-designateclient)](https://pypi.python.org/pypi/python-designateclient/2.1.0), and URLs.


hroncok:
Dos this have tests?

mcyprian:
Good point, I will write some.

hroncok:
Good to merge

PR #56, opened 2016-06-23
License files

mcyprian:
Doc files are separated to license files and other doc files, License files are listed in %license section of the spec file instead of %doc.
This PR also contains minor fix of created spec file's name when -s/--srpm option is set.

### DIFF
--- a/pyp2rpm/metadata_extractors.py
+++ b/pyp2rpm/metadata_extractors.py
@@ -354,8 +354,9 @@ class SetupPyMetadataExtractor(LocalMetadataExtractor):
             else settings.DEFAULT_PYTHON_VERSION
         archive_data['python_versions'] = py_vers[1:] if py_vers \
             else [settings.DEFAULT_ADDITIONAL_VERSION]
-
-        archive_data['doc_files'] = self.doc_files
+        doc_files = self.doc_files
+        archive_data['doc_files'] = [doc for doc in doc_files if 'license' not in doc.lower()]
+        archive_data['doc_license'] = [doc for doc in doc_files if 'license' in doc.lower()]
         archive_data['py_modules'] = self.py_modules
         archive_data['has_test_suite'] = self.has_test_suite
         archive_data['has_bundled_egg_info'] = self.has_bundled_egg_info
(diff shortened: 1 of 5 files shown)


hroncok:
More tests with different files? COPYRIGHT etc.?

mcyprian:
Are there any other file names except */LICENSE */COPYRIGHT */COPYING that should be moved to %license?

hroncok:
well, see for example https://github.com/encukou/py3c/blob/master/LICENSE.MIT https://github.com/encukou/py3c/blob/master/doc/LICENSE.CC-BY-SA-3.0 and similar.

pyp2rpm/metadata_extractors.py
```diff
@@ -550,8 +552,11 @@ def data_from_archive(self):
         archive_data['summary'] = self.summary
         archive_data['home_page'] = self.home_page
         doc_files = self.doc_files
-        archive_data['doc_files'] = [doc for doc in doc_files if 'license' not in doc.lower()]
-        archive_data['doc_license'] = [doc for doc in doc_files if 'license' in doc.lower()]
+        archive_data['doc_files'] = [doc for doc in doc_files if all(s not in doc.lower() for s in
+                                                                     settings.LICENSE_FILES)]
+        archive_data['doc_license'] = [doc for doc in doc_files if any(s in doc.lower() for s in
+                                                                       settings.LICENSE_FILES)]
```
hroncok: I read this two times. Better to have a function?

mcyprian:
Fixed.

pyp2rpm/metadata_extractors.py
```diff
@@ -355,7 +363,8 @@ def data_from_archive(self):
         archive_data['python_versions'] = py_vers[1:] if py_vers \
             else [settings.DEFAULT_ADDITIONAL_VERSION]
 
-        archive_data['doc_files'] = self.doc_files
+        archive_data['doc_files'], archive_data[
+            'doc_license'] = self.separate_license_files(self.doc_files)
```
hroncok: Any chance of some more readable line break? Not necessarily, but this feel, ehm, ugly?

hroncok:
Everything except the cosmetic comment about the line break seems OK.

mcyprian:
I think it looks a bit better now :) thanks for the review @hroncok.

PR #58, opened 2016-06-27
no need for %% escape. this is not .spec file

glensc:
no need for %% escape. this is not .spec file :)


mcyprian:
You are right, thanks

PR #64, opened 2016-07-04
update pld template

glensc:
does much better job than previous template.

btw, i added some local macros as they are pld specific, and was wondering why some of the macros are defined in sourcecode [pyp2rpm/filters.py](https://github.com/fedora-python/pyp2rpm/blob/master/pyp2rpm/filters.py) and the others in [pyp2rpm/templates/macros.spec](https://github.com/fedora-python/pyp2rpm/blob/master/pyp2rpm/templates/macros.spec)?


mcyprian:
[Filters](http://jinja.pocoo.org/docs/dev/templates/#filters) and [macros](http://jinja.pocoo.org/docs/dev/templates/#macros) are not the same thing. It is described properly in the Jinja documentation.

PR #71, opened 2016-08-09
Tests: Run mybin.py using 'python' and not the shebang

hroncok:
Otherwise /usr/bin/python3 is used and not the one from virtualenv/tox

Fixes travis-ci/travis-ci#6168


mcyprian:
Thanks @hroncok.

Issue #70, opened 2016-08-09
ValueError: unknown url type: 'UNKNOWN'


frenzymadness:
When I try to use pyp2rpm on Pymacs package I get following error: ValueError: unknown url type: 'UNKNOWN'

In get_url function in package_getters.py release_urls variable is empty list and release_data contains this:

```
{'platform': 'UNKNOWN', 'requires_python': None, 'cheesecake_installability_id': None, 'author': 'François Pinard', 'package_url': 'http://pypi.python.org/pypi/Pymacs', 'release_url': 'http://pypi.python.org/pypi/Pymacs/0.25', 'classifiers': [], 'home_page': 'UNKNOWN', 'maintainer_email': None, 'keywords': None, 'cheesecake_code_kwalitee_id': None, '_pypi_ordering': 0, 'description': 'False', **'download_url': 'UNKNOWN'**, 'maintainer': None, 'version': '0.25', 'license': 'GPLv2', 'cheesecake_documentation_id': None, 'docs_url': None, 'name': 'Pymacs', 'bugtrack_url': None, 'summary': 'Interface between Emacs Lisp and Python', 'author_email': 'pinard@iro.umontreal.ca', '_pypi_hidden': False, 'downloads': {'last_day': 0, 'last_month': 0, 'last_week': 0}}
```

This data comes from

```
        release_urls = client.release_urls(name, version)
        release_data = client.release_data(name, version)
```

But I cannot find definition of client.release_[data|urls] so I cannot go more to deep.

pyp2rpm version = '3.1.2'

frenzymadness:
Original traceback:

```
Traceback (most recent call last):
  File "/home/lbalhar/.virtualenvs/py3-packaging-porting/bin/pyp2rpm", line 11, in <module>
    sys.exit(main())
  File "/home/lbalhar/.virtualenvs/py3-packaging-porting/lib/python3.5/site-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/home/lbalhar/.virtualenvs/py3-packaging-porting/lib/python3.5/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/home/lbalhar/.virtualenvs/py3-packaging-porting/lib/python3.5/site-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
[...]

PR #75, opened 2016-08-15
Added handling of 'UNKNOWN' URL when package has no sources.

frenzymadness:


hroncok:
Reference #70 in the commit message?

frenzymadness:
Yes, I forget to mention related issue in commit message but I can close it manually after merge.

mcyprian:
The only thing that is changed by this fix is that 'PACKAGE WITHOUT SOURCES' appears at the end of Traceback instead of 'UNKNOWN'. It is not possible to generate the spec file without package sources. I suggest writing error message to stderr instead of log and raising SystemExit in both cases - url is None or set to 'UNKNOWN'.

frenzymadness:
Ok, I am gonna rearrange this part.

What is preferred way to print to stderr? I prefer to use `print("error", file=sys.stderr)` but this brings dependency on future module for Python2. Another possibility is to use only `sys.stderr.write("error\n")`.

mcyprian:
The second option is used in pyp2rpm's sources several times already, so keep it consistent please.

frenzymadness:
I've made some changes as you suggested. Also, I created new parametrized test for testing exceptions in get_url function.

Issue #72, opened 2016-08-11
inconsistent appearance of "packages" key in metadata


mcyprian:
`pyp2rpm simpleeval` (venv extraction is enabled):
'packages': set() appears in the metadata object

`pyp2rpm --no-venv simpleeval` (venv deactivated):
key "packages" is not included in the metadata and packages_data's __getattr__ method returns 'TODO:' when attribute packages of the object is used.

PR #77, opened 2016-08-31
Refactoring of metadata extractors

mcyprian:
In this PR:
- Module extract_distribution replaced by direct subclass of distutils.Command extract_dist
- Much more simple and readable code, the same performance of metadata extraction
- Old metadata extractor parsing setup.py script and parsing methods from archive module removed
- Common interface of SetupPyMetadataExtractor and WheelMetadataExtractor - one data_from_archive method can be used for both
- Venv metadata extraction rewritten as an decorator
- Complex tests for all metadata attributes
- Default empty values in package_data


mcyprian:
If runpy fails to run extract_dist command (Syntax or Runtime error occurs), command is run using external python2 interpreter with --output option. Extracted metadata are sent to standard output in json format, read and deserialized in metadata extractor.

At this moment string "extracted json data:\n" is at the beginning of json metadata, because other messages from setup.py script or command can appear on stdout. Maybe some more sophisticated separator could be used.

hroncok:
Does it also work the other way around (as described [here](https://github.com/fedora-python/pyp2rpm/issues/80#issuecomment-244990059))?

mcyprian:
No it doesn't. I was focused at the original issue #72.  Do you suggest to try running python3 in the subprocess in this case or can we expect that package is python2 incompatible and fail? At the moment when ImportError occurs we don't have any metadata of the package. I am not sure if there is deterministic way to find out if package is python2 compatible or not (classifiers are usually not very confident) even if we had needed metadata.

hroncok:
Simply try _the other_ interpreter. (I.e. 2 on 3 and 3 on 2).

mcyprian:
Now it works for both cases.

PR #84, opened 2016-09-29
split build deps + test build deps

glensc:
preserves backward compatibility with distros whose templates are present
except that uniqueness may be lost.

altho adding unique filter to template should be easy, but i don't know the language to add.

### DIFF
--- a/pyp2rpm/metadata_extractors.py
+++ b/pyp2rpm/metadata_extractors.py
@@ -100,6 +100,7 @@ class LocalMetadataExtractor(object):
         self.venv = venv
         self.base_python_version = base_python_version
         self.metadata_extension = metadata_extension
+        self.builtin_build_deps = [['BuildRequires', 'python2-devel'], ['BuildRequires', 'python-setuptools']]
 
     def name_convert_deps_list(self, deps_list):
         for dep in deps_list:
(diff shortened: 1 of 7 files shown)


glensc:
it may also call for refactor need:

i looked the extractors code closely then pyp2rpm renamed metadata key named "test" to "build", meaning the existing `build_deps` property should had been actually named as `test_deps`.

mcyprian:
Test requires are included in build requires now, but there are also true build requires in the list returned from build_deps/build_deps_from_setup_py. If it is really useful to keep them separate for pld template, I suggest to do it in code of build_deps property. 
I am currently working on [refactoring of extract_dist commad and metadata_extracors](https://github.com/fedora-python/pyp2rpm/pull/77), Appending of test requires to build_deps is more visible [there](https://github.com/fedora-python/pyp2rpm/commit/5f89ba3654f485ba89e5f8e7d1498a48a1aaf1ad#diff-b03f151d580cccef91f5d2c92a70977aR270) (It was done in extraction command before). It would be better to apply changes to new code of metadata_extractors.

glensc:
i would like to see test and build deps separately like i've shown for pld template.

it's not really possible to distinguish whether test require is actually needed to only build. they put weird stuff in their setup.py :) so it's always up to packager to verify the dependencies. if they get listed in side `%if tests`, it's easier to move them outside `%if tests`

anyway, i'll revisit this pr once your previous code gets merged. this was anyway more of proof of concept to show what is i want and get some feedback.

PR #89, opened 2016-11-14
Base Python version (`-b`) in metadata extraction

### LINKED ISSUE #90: Use python2 to run extraction command when -b2 option is specified
We might consider using python2 as a primary interpreter to run extraction command when -b2 option is used. This could help us to get more accurate metadata.
In the other hand sys.executable always works, but there is no guarantee that python2 is present on user's machine/virtualenvironment.

irushchyshyn:
Use the base Python version (if provided with `-b` option) to extract metadata by running `setup.py` for the package.

Fixes #90.


mcyprian:
Everything seems to work fine except this case:
I have pyp2rpm installed in virtualenvironment, I run command "pyp2rpm -b2 package"
and extraction of metadata using python2 interpreter fails for some reason (e.g. python2 is not installed). After this happens pyp2rpm uses /usr/bin/python3 in second attempt to extract data instead of python3 inside virtualenv (/home/mcyprian/.virtualenvs/pyp2rpm/bin/python3).
You can reference issue #90 in your commit.

irushchyshyn:
@mcyprian Thank you for the comment.
I have fixed the issue you have mentioned in e0b0b65. Now if the version supplied with -b option or the alternative version is a current executable, the executable will be used.

mcyprian:
Approved.

PR #91, opened 2016-11-15
Drop Python 2.6 support and remove py26 from tox envlist

irushchyshyn:
Running tests with tox has a dependency on `virtualenv-api` which stopped supporting Python 2.6 a long time ago ([commit](https://github.com/sjkingo/virtualenv-api/commit/e2aaad332d383b1673159dedb666bd70d487701b)) . With the changes in the latest `virtualenv-api` release 2.1.13, `pyp2rpm`'s tests started to reflect this and running tests in py26 tox environment started to fail. Continuing to support Python 2.6 officially without being able to run test in py26 makes no sense. Unless we want to introduce dependency on a specific `virtualenv-api` version as [here](https://github.com/fedora-python/pyp2rpm/pull/89/commits/5b28f392524f1280b97455ea6a218ada232df201).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 
 python:
-- '2.6'
 - '2.7'
 - '3.4'
 - '3.5'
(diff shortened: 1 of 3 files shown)


mcyprian:
Looks good, thank you

PR #92, opened 2016-11-23
 Fix "File not found" errors

mcyprian:
- Improvement of packages and modules
    - dash in pkg name not replaced with underscore (virtualenv module extracts name in the correct form)
    - PyPI packages containing modules only
      
- Impovement of .pth files detection
    - packages without '.' in name containing .pth file (PasteDeploy)
    - packages containing '.' in name without .pth files (giturl.py)
    - py2 specific .pth files not added to python3 subpackage (zc.buildout)
- Marker of spec file test
- Encoding in mybin.py script

### DIFF
--- a/pyp2rpm/metadata_extractors.py
+++ b/pyp2rpm/metadata_extractors.py
@@ -135,7 +135,10 @@ class LocalMetadataExtractor(object):
     @property
     def has_pth(self):
         """Figure out if package has pth file """
-        return "." in self.name
+        if virtualenv is None:
+            return "." in self.name
+        else:
+            return None
 
     @property
     def has_extension(self):
--- a/pyp2rpm/metadata_extractors.py
+++ b/pyp2rpm/metadata_extractors.py
@@ -85,7 +85,7 @@ def venv_metadata_extension(extraction_fce):
 
 
 def process_description(description_fce):
-    """Removes special character delimiters, titles 
+    """Removes special character delimiters, titles
     and wraps paragraphs.
     """
     def inner(description):
@@ -165,8 +165,10 @@ class LocalMetadataExtractor(object):
             setattr(data, "scripts", utils.remove_major_minor_suffix(data.data['scripts']))
         # for example nose has attribute `packages` but instead of name listing the pacakges
         # is using function to find them, that makes data.packages an empty set
-        if getattr(data, "packages") == set():
+        # if virtualenv is disabled
+        if virtualenv is None and getattr(data, "packages") == set():
             data.packages = set([data.name])
+
 
         return data
 
--- a/pyp2rpm/templates/fedora.spec
+++ b/pyp2rpm/templates/fedora.spec
@@ -88,7 +88,7 @@ ln -sf %{_bindir}/{{ script|script_name_for_python_version(pv) }} %{buildroot}/%
 {%- endfor %}
 {%- if data.py_modules %}
 {% for module in data.py_modules -%}
-{%- if pv == '3' -%}
+{%- if pv == '3' %}
 {{ '%{python2_sitelib}'|sitedir_for_python_version(pv) }}/__pycache__/*
 {%- endif %}
 {{ '%{python2_sitelib}'|sitedir_for_python_version(pv) }}/{{ data.name | module_to_path(module) }}.py{% if pv != '3'%}*{% endif %}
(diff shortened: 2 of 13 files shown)


tests/test_virtualenv.py
```diff
@@ -66,13 +66,18 @@ def teardown_method(self, method):
...
     ])
     def test_get_dirs_differance(self, bin_diff, package_diff, expected):
         flexmock(DirsContent).should_receive('__sub__').and_return(
             DirsContent(bin_diff, package_diff))
-        assert self.venv.get_dirs_differance == expected
+        self.venv.get_dirs_differance()
+        assert ((self.venv.data['packages'],
+                 self.venv.data['py_modules'],
+                 self.venv.data['scripts'],
+                 self.venv.data['has_pth']) == expected)
```
irushchyshyn, line 84: Can we also test `has_pth` when there are .pth files?

pyp2rpm/templates/fedora.spec
```diff
@@ -88,22 +88,26 @@ ln -sf %{_bindir}/{{ script|script_name_for_python_version(pv) }} %{buildroot}/%
...
 {{ '%{python2_sitelib}'|sitedir_for_python_version(pv) }}/__pycache__/*
 {%- endif %}
 {{ '%{python2_sitelib}'|sitedir_for_python_version(pv) }}/{{ data.name | module_to_path(module) }}.py{% if pv != '3'%}*{% endif %}
 {%- endfor %}
 {%- endif %}
 {%- if data.has_extension %}
-{{ '%{python2_sitearch}'|sitedir_for_python_version(pv) }}/{{ data.name | module_to_path(data.underscored_name) }}
+{%- if data.has_packages %}
+{%- for package in data.packages %}
+{{ '%{python2_sitearch}'|sitedir_for_python_version(pv) }}/{{ package | package_to_path(data.name) }}
```
irushchyshyn, line 100: Why do we supply `data.underscored_name` to `package_to_path` in the rest of the templates, and `data.name` only for fedora.spec?

Also, it would be nice to have tests for this kind of spec files with extensions to make it more clear how they are affected.

pyp2rpm/metadata_extractors.py
```diff
@@ -165,8 +168,10 @@ def extract_data(self):
             setattr(data, "scripts", utils.remove_major_minor_suffix(data.data['scripts']))
         # for example nose has attribute `packages` but instead of name listing the pacakges
         # is using function to find them, that makes data.packages an empty set
-        if getattr(data, "packages") == set():
-            data.packages = set([data.name])
+        # if virtualenv is disabled
+        if virtualenv is None and getattr(data, "packages") == set():
+              data.packages = set([data.name])
```
irushchyshyn: Nitpicking: The indentation seems wrong.

pyp2rpm/metadata_extractors.py
```diff
@@ -135,7 +135,10 @@ def name_convert_deps_list(self, deps_list):
     @property
     def has_pth(self):
         """Figure out if package has pth file """
-        return "." in self.name
+        if virtualenv is None:
+            return "." in self.name
+        else:
```
irushchyshyn, line 140: Nitpicking:  The `else` statement is redundant in this case.

mcyprian:
Thanks for the review @irushchyshyn, I have fixed all issues mention above.

pyp2rpm/metadata_extractors.py.orig
```diff
@@ -0,0 +1,500 @@
+import logging
```
irushchyshyn: This file must have been committed by mistake.

irushchyshyn (commented):
@mcyprian Thanks for the changes.
Looks good to me.  Except for the [ pyp2rpm/metadata_extractors.py.orig](https://github.com/fedora-python/pyp2rpm/pull/92/files#diff-52552953033b9520a50eb09618af3934) which seems not needed.

pyp2rpm/metadata_extractors.py
mcyprian: You are right of course, but I think it is more obvious what return values we expect in other modules when it is there.

irushchyshyn: Fair enough.

irushchyshyn:
:+1: 
Not sure though if I am entitled to merge.

mcyprian:
Yes, you are.

irushchyshyn:
Merged! Thanks.

Issue #87, opened 2016-11-08
package rename option is ignored - importing Sphinx package with wrong name


ssbarnea:
While trying to convert one of the most popular python packages `Sphinx` I found this bug.

    pyp2rpm -r python-sphinx -d . --srpm Sphinx

Instead of generating a rpm for the package named `python-sphinx` it will generate `python-Sphinx`. Both filename and NAME-VERSION-RELEASE do contain `python-Sphinx` instead of `python-Sphinx`.

Base on Fedora naming convention I would even assume that the package name would be lowercases anyway, but that's not the case. 

I would not mind making a pull request but first I want to understand which approach for fixing this would be preferred.

mcyprian:
We should take a loot at #66 before fixing this issue. It is also related to pypi_name macro usage in the template.

Issue #66, opened 2016-07-20
Errors with packages named python-*


hroncok:
Take a package that is called python-something (on Python level, i.e. on PyPI) and run it trough pyp2rpm, tak for example python-stdnum, this is the current output (shortened where not relevant):

```
%global pypi_name python-stdnum

Name:           python-stdnum

Source0:        https://files.pythonhosted.org/packages/source/p/%{pypi_name}/%{pypi_name}-%{version}.tar.gz

%package -n     python2-stdnum
%{?python_provide:%python_provide python2-%{pypi_name}}
%description -n python2-stdnum

%package -n     python3-stdnum
%{?python_provide:%python_provide python3-%{pypi_name}}
%description -n python3-stdnum

%package -n python-%{pypi_name}-doc
%description -n python-%{pypi_name}-doc

%files -n python2-stdnum

%files -n python3-stdnum

%files -n python-%{pypi_name}-doc
```

The important thing here is that pyp2rpm somewhere knows the RPM subpackages should be pythonX-stdnum and not pythonX-python-stdnum, but it does so without honoring macros and does so not in the `%python_provide` macros, resulting in following provides:

**python2-stdnum:**
- python-python-stdnum
- python2-stdnum

**python3-stdnum:**
- python3-stdnum

Also, note the python-python-stdnum-doc package.

The correct resolution here would be to set `%pypi_name` just to `stdnum` and change the bits where needed to be chnaged. However, that would be a bit false, because the PyPI name is not just `stdnum`, so maybe rename the macro?

This is the proposed spec:

```
%global srcname stdnum

Name:           python-%{srcname}

Source0:        https://files.pythonhosted.org/packages/source/p/%{name}/%{name}-%{version}.tar.gz

%package -n     python2-%{srcname}
%{?python_provide:%python_provide python2-%{srcname}}
%description -n python2-%{srcname}

%package -n     python3-%{srcname}
%{?python_provide:%python_provide python3-%{srcname}}
%description -n python3-%{srcname}

%package -n python-%{srcname}-doc
%description -n python-%{srcname}-doc

%files -n python2-%{srcname}

%files -n python3-%{srcname}

%files -n python-%{srcname}-doc
```

mcyprian:
We could use **macroed_pkg_name** filter that is used to generate global name and names of the packages. Result for normal packages would be unchanged and
`%{?python_provide:%python_provide python3-stdnum}` would appear there in corner cases like python-stdnum.

[...]

PR #94, opened 2016-12-05
Make `-r` option work (#87)

irushchyshyn:
- Introduce `%{srcname}` macro to be used alongside `%{pypi_name}` if the  package name is to be changed (applies also for packages already starting with `python-` or `Python-` in PyPI)
- Fix filters tests to work with the new signature for `macroed_pkg_name`
- Add a new integration test for `python-sphinx.spec`
- Corner case: fix searching for Sphinx directory, when it is tested itself (another `doc/conf.py` in tests folder)

Fixes issues #87 and #66.

### DIFF
--- a/pyp2rpm/templates/epel7.spec
+++ b/pyp2rpm/templates/epel7.spec
@@ -78,7 +78,7 @@ ln -sf %{_bindir}/{{ script|script_name_for_python_version(pv) }} %{buildroot}/%
 {%- endfor %}
 {%- endif %}
 {% for pv in [data.base_python_version] + data.python_versions %}
-%files -n {{ data.pkg_name|macroed_pkg_name(data.srcname)|name_for_python_version(pv, True, True) }} 
+%files -n {{ data.pkg_name|macroed_pkg_name(data.srcname)|name_for_python_version(pv, True, True) }}
 %doc {{data.doc_files|join(' ') }}
 {%- for script in data.scripts %}
 {%- if pv == data.base_python_version %}
--- a/pyp2rpm/templates/epel7.spec
+++ b/pyp2rpm/templates/epel7.spec
@@ -1,8 +1,11 @@
 {{ data.credit_line }}
 {% from 'macros.spec' import dependencies, for_python_versions, underscored_or_pypi -%}
 %global pypi_name {{ data.name }}
+{%- if data.srcname %}
+%global srcname {{ data.srcname }}
+{%- endif %}
 
-Name:           {{ data.pkg_name|macroed_pkg_name(data.name) }}
+Name:           {{ data.pkg_name|macroed_pkg_name(data.srcname) }}
 Version:        {{ data.version }}
 Release:        1%{?dist}
 Summary:        {{ data.summary }}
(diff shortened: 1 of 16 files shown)


pyp2rpm/templates/epel7.spec
```diff
@@ -75,7 +78,7 @@ ln -sf %{_bindir}/{{ script|script_name_for_python_version(pv) }} %{buildroot}/%
 {%- endfor %}
 {%- endif %}
 {% for pv in [data.base_python_version] + data.python_versions %}
-%files -n {{ data.pkg_name|macroed_pkg_name(data.name)|name_for_python_version(pv, True, True) }} 
+%files -n {{ data.pkg_name|macroed_pkg_name(data.srcname)|name_for_python_version(pv, True, True) }}
```
mcyprian: Trailing whitespace present at the end of this line.

pyp2rpm/templates/fedora.spec
```diff
@@ -74,7 +77,7 @@ ln -sf %{_bindir}/{{ script|script_name_for_python_version(pv) }} %{buildroot}/%
 {%- endfor %}
 {%- endif %}
 {% for pv in [data.base_python_version] + data.python_versions %}
-%files -n {{ data.pkg_name|macroed_pkg_name(data.name)|name_for_python_version(pv, True) }} 
+%files -n {{ data.pkg_name|macroed_pkg_name(data.srcname)|name_for_python_version(pv, True) }}
```
mcyprian: Trailing whitespace present at the end of this line.

mcyprian (commented):
This PR fixes also  #66, right? Could you reference it in one of the commits? 
It seems to work fine for both cases:
    - name of the package is explicitly changed by -r option
    - original PyPI name contains python- prefix

mcyprian:
There is also couple of packages containing "Python-" prefix on PyPI. Python-Crest for example, in this case result is following:
```
%global pypi_name Python-Crest

Name:           python-%{pypi_name}
```
Maybe we should cover this case as well.

pyp2rpm/templates/epel7.spec
irushchyshyn: Fixed in 2c7783f710a97c4e8dbb1bf29ce35327aaea86c0

pyp2rpm/templates/fedora.spec
irushchyshyn: Fixed in 2c7783f710a97c4e8dbb1bf29ce35327aaea86c0

irushchyshyn:
> There is also couple of packages containing "Python-" prefix on PyPI. Python-Crest for example, in this case result is following:
> %global pypi_name Python-Crest
> Name:                    python-%{pypi_name}
> Maybe we should cover this case as well.

Thanks for a suggestion. I have covered those packages in 2c7783f

irushchyshyn:
> This PR fixes also #66, right? Could you reference it in one of the commits?
> It seems to work fine for both cases:
> - name of the package is explicitly changed by -r option
> - original PyPI name contains python- prefix

Thanks for a review. I have referenced the issue in the PR description.

mcyprian:
Works like a charm :)

PR #98, opened 2016-12-19
Fix locating `setup.py` (Issue #95)

### LINKED ISSUE #95: setup.py not found
Running `pyp2rpm ndg-httpsclient` fails with an error: "setup.py not found, maybe local_file is not proper source archive." Downloaded archive ndg_httpsclient-0.4.2.tar.gz seems to be alright and contains setup.py script.

irushchyshyn:
The problem with the listed packages is that the archive name differs from the PyPI name.

Take [ndg-httpsclient](https://pypi.python.org/pypi/ndg-httpsclient). The downloaded tar is `ndg_httpsclient-0.4.2.tar.gz`. As we solely depend on the PyPI name, we assume that the package will be extracted into `ndg-httpsclient-0.4.2`, which causes errors when locating `setup.py` and generating a spec file. Also the build later fails in `%prep` section in `%autosetup` where we also depend on PyPi name.

* fix matching `setup.py` file
* fix `%autosetup` to change into the unpacked directory

Tested the fixes in copr on the set of packages reported in the issue (and a couple of other packageswith the same issue, found during testing): https://copr.fedorainfracloud.org/coprs/ishcherb/pypi_builds1/packages/

Fixes #95.

### DIFF
--- a/pyp2rpm/templates/fedora.spec
+++ b/pyp2rpm/templates/fedora.spec
@@ -40,7 +40,7 @@ Documentation for {{ data.name }}
 {%- endif %}
 
 %prep
-%autosetup -n %{pypi_name}-%{version}
+%autosetup -n {{ data.dirname|replace(data.name, '%{pypi_name}')|replace(data.version, '%{version}')|default('%{pypi_name}-%{version}', true) }}
 {%- if data.has_bundled_egg_info %}
 # Remove bundled egg-info
 rm -rf %{pypi_name}.egg-info
--- a/pyp2rpm/archive.py
+++ b/pyp2rpm/archive.py
@@ -55,6 +55,9 @@ class ZipWrapper(object):
     def getmembers(self, *args, **kwargs):
         return self._wrapped_obj.infolist(*args, **kwargs)
 
+    def getnames(self, *args, **kwargs):
+        return self._wrapped_obj.namelist(*args, **kwargs)
+
     def extractfile(self, *args, **kwargs):
         return self._wrapped_obj.open(*args, **kwargs)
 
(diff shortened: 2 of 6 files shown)


pyp2rpm/templates/fedora.spec
```diff
@@ -40,7 +40,7 @@ Documentation for {{ data.name }}
 {%- endif %}
 
 %prep
-%autosetup -n %{pypi_name}-%{version}
+%autosetup -n {{ data.dirname|replace(data.name, '%{pypi_name}')|replace(data.version, '%{version}')|default('%{pypi_name}-%{version}', true) }}
```
mcyprian, line 43: This change should be applied to epel7.spec template as well.

mcyprian (commented):
Main issue is fixed, good job :) I think using data.dirname instead of data.underscored_name in %files section (.pth, .egg-info files) might also fix cases like [BioPandas File not found error](https://copr-be.cloud.fedoraproject.org/results/ishcherb/pypi_builds1/fedora-26-x86_64/00491050-python-BioPandas/build.log.gz).

pyp2rpm/templates/fedora.spec
irushchyshyn: Applied in 66a5e79

irushchyshyn:
Thanks for the review @mcyprian !
> I think using data.dirname instead of data.underscored_name in %files section (.pth, .egg-info files) might also fix cases like BioPandas File not found error.

I do not think we should use `dirname` in %files. It might fix the error for BioPandas, but the .egg-info files are not always named the same way the tar file is. They are more often underscored I would say.

mcyprian:
Yes you're right, in that case virtualenv module could hopefully help here but that is different issue...

PR #99, opened 2016-12-22
Fix IndexError in versions extraction (Issue #96)

### LINKED ISSUE #96: IndexError in versions extraction
Running `pyp2rpm imread` fails with following traceback:

```
File "/home/mcyprian/Codes/devel/pyp2rpm/pyp2rpm/metadata_extractors.py", line 427, in data_from_archive
    archive_data = super(SetupPyMetadataExtractor, self).data_from_archive
  File "/home/mcyprian/Codes/devel/pyp2rpm/pyp2rpm/metadata_extractors.py", line 221, in data_from_archive
    archive_data['python_versions']) = self.versions_from_archive
  File "/home/mcyprian/Codes/devel/pyp2rpm/pyp2rpm/metadata_extractors.py", line 371, in versions_from_archive
    base_python_version = python_versions[0]
IndexError: list index out of range

```

irushchyshyn:
Fixes #96.

### DIFF
--- a/pyp2rpm/metadata_extractors.py
+++ b/pyp2rpm/metadata_extractors.py
@@ -366,16 +366,21 @@ class SetupPyMetadataExtractor(LocalMetadataExtractor):
 
     @property
     def versions_from_archive(self):
+        """Return Python versions extracted from trove classifiers or default.
+        Returns:
+            (str, list) base_python_version, python_versions
+        """
         py_vers = utils.versions_from_trove(self.metadata['classifiers'])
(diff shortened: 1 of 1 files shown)


mcyprian (commented):
I am still getting IndexError but the reason for this is that /usr/bin/python which is executed after first extraction failure can import numpy on my local machine. 
I think it is not necessary to fix anything here, maybe we could end with nicer error message in this case and ask upstream to remove numpy import from setup.py if possible.

irushchyshyn:
> I am still getting IndexError but the reason for this is that /usr/bin/python which is executed after first extraction failure can import numpy on my local machine.

Thanks for pointing this out, fixed now.

PR #101, opened 2017-02-03
Fix converting corner cases like numpy (Issue #97)

### LINKED ISSUE #97: Function in metadata
Running `pyp2rpm numpy` fails with following exception:
```
Traceback (most recent call last):
  File "setup.py", line 391, in <module>
    setup_package()
  File "setup.py", line 383, in setup_package
    setup(**metadata)
  File "/tmp/tmpqe85xcha/numpy-1.12.0b1/numpy/distutils/core.py", line 169, in setup
    return old_setup(**new_attr)
  File "/usr/lib64/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/usr/lib64/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/usr/lib64/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/home/mcyprian/Codes/devel/pyp2rpm/command/extract_dist.py", line 51, in run
    sys.stdout.write("extracted json data:\n" + json.dumps(self.metadata))
  File "/usr/lib64/python2.7/json/__init__.py", line 244, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib64/python2.7/json/encoder.py", line 207, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib64/python2.7/json/encoder.py", line 270, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib64/python2.7/json/encoder.py", line 184, in default
    raise TypeError(repr(o) + " is not JSON serializable")
TypeError: <function generate_config_py at 0x7f9a5ce81230> is not JSON serializable
```
It seems function appears in metadata we try to serialize in extract_dist command. I think this is not very frequent case, but we should take a look.

irushchyshyn:
Use default serializer when sending extracted metadata to stdout to convert functions and other non serializable objects to string rather than raise an exception (fixes #97). Provide unit tests. 

The PR also includes:
* Fix for test failures after various dependencies updated (jinja, setuptools, pytest-runner)
Travis Ci Cron Job to identify these in the future?

* Run tests in Python 3.6 and recent dev version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ python:
 - '2.7'
 - '3.4'
 - '3.5'
+- '3.6'
+- 'nightly'
 
 cache:
   directories:
--- a/command/extract_dist.py
+++ b/command/extract_dist.py
@@ -48,7 +48,8 @@ class extract_dist(Command):
         is specified, assigns metadata dictionary to class_metadata variable otherwise.
         """
         if self.stdout:
-            sys.stdout.write("extracted json data:\n" + json.dumps(self.metadata))
+            sys.stdout.write("extracted json data:\n" + json.dumps(
+                self.metadata, default=to_str))
         else:
             extract_dist.class_metadata = self.metadata
 
@@ -69,6 +70,7 @@ def to_list(var):
 
 def to_str(var):
     """Similar to to_list function, but for string attributes."""
-    if not isinstance(var, str):
-        return 'TODO'
-    return var
+    try:
+        return str(var)
+    except TypeError:
+        raise ValueError("{} cannot be converted to string.".format(var))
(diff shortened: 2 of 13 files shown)


.travis.yml
```diff
@@ -4,6 +4,8 @@ python:
 - '2.7'
 - '3.4'
 - '3.5'
+- '3.6'
```
mcyprian, line 7: Good idea to add this :)

mcyprian (approved):
Interesting and simple solution, looks good to me. I created Travis CI Cron job to run tests each week.

Issue #100, opened 2017-01-19
Integration with spec2scl


irushchyshyn:
Integrate [spec2scl](https://github.com/sclorg/spec2scl/) to enable the user to create an SCL-style spec file for a PyPI package.

There should be an option (e.g. `--sclize`) to convert tags and macro definitions and produce a Software Collection spec file. The user should be able to provide optional arguments defined for `spec2scl`.
The implementation will require introducing a dependency on  [spec2scl](https://github.com/sclorg/spec2scl/).

hroncok:
Note: It can be optional dependency and the feature will simply not be available if spec2scl is not installed.

irushchyshyn:
That is a very good suggestion, thanks @hroncok

PR #102, opened 2017-02-15
Provide --sclize option

irushchyshyn:
Enable generating sclized spec files using spec2scl.
The following spec2scl options can be provided alongside:

- --no-meta-runtime-dep
- --no-meta-buildtime-dep
- --skip-functions
- --no-deps-convert
- --list-file

Issue #100

### DIFF
--- a/pyp2rpm/bin.py
+++ b/pyp2rpm/bin.py
@@ -10,14 +10,64 @@ from pyp2rpm.logger import register_file_log_handler, register_console_log_handl
 import click
 try:
     from spec2scl.convertor import Convertor as SclConvertor
-except:
+except ImportError:
     SclConvertor = None
+except Exception as err:
+    SclConvertor = None
+    click.echo('Warning: An unexpected error occured when trying'
+               ' to import spec2scl: {}'.format(err))
 
 
 CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
 
 
-@click.command(context_settings=CONTEXT_SETTINGS)
+class Pyp2rpmCommand(click.Command):
+    """Base class for pyp2rpm command.
+
+    Modify the help message for SCL related options,
+    to single them out in a separate group.
+    """
+
+    def format_options(self, ctx, formatter):
+        """Writes SCL related options into the formatter as a separate group."""
+        super(Pyp2rpmCommand, self).format_options(ctx, formatter)
+        scl_opts = []
+        for param in self.get_params(ctx):
+            if isinstance(param, SclizeOption):
+                scl_opts.append(param.get_scl_help_record(ctx))
+
+        if scl_opts:
+            with formatter.section('SCL related options'):
+                formatter.write_dl(scl_opts)
+
+
+class SclizeOption(click.Option):
+    """Base class for SCL related options.
+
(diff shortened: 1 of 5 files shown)


pyp2rpm/bin.py
```diff
@@ -8,6 +8,10 @@
 from pyp2rpm.logger import register_file_log_handler, register_console_log_handler
 
 import click
+try:
+    from spec2scl.convertor import Convertor as SclConvertor
+except:
```
mcyprian: Can something except ImportError (ModuleNotFoundError) occur here? Maybe we could log the information when something unexpected happens.

pyp2rpm/bin.py
```diff
@@ -98,6 +126,11 @@ def main(package, v, d, s, r, proxy, srpm, p, b, o, t, venv):
     converted = convertor.convert()
     logger.debug('Convertor: {0} succesfully converted.'.format(convertor))
 
+    if sclize:
+        converted = convert_to_scl(converted, scl_kwargs)
+        if not converted:
+            return
```
mcyprian: I caught the idea of this part, but at the beginning it was not clear to me in which cases is the return value of convert_to_scl None/Empty string. Can you add a short comment please.

mcyprian (changes requested):
Simple and elegant solution, a few things could be improved in my opinion:
- when running pyp2rpm package --no-meta-runtime-dep it seems this option is simply ignored. Combining these options with --sclize should be required or user should be warn that option has no effect in this case.
- I know that help is autogenerated in click, but isn't it possible to make the fact that the last 4 options are SCL specific a bit more obvious in the help message?
- pyp2rpm's man page (file pyp2rpm.1 in the main project directory) should be updated as well. It should be possible to generate man page from the help message directly, but I've never tried how it works https://www.gnu.org/software/help2man/

irushchyshyn:
Thank you for the review @mcyprian ! 

>     when running pyp2rpm package --no-meta-runtime-dep it seems this option is simply ignored. Combining these options with --sclize should be required or user should be warn that option has no effect in this case.
I have added a validation for a case like that.

>     I know that help is autogenerated in click, but isn't it possible to make the fact that the last 4 options are SCL specific a bit more obvious in the help message?
I have modified help formatting for options, to put SCL related options in a separate options group. Please let me know what you think and how it looks now.

    pyp2rpm's man page (file pyp2rpm.1 in the main project directory) should be updated as well. It should be possible to generate man page from the help message directly, but I've never tried how it works https://www.gnu.org/software/help2man/
Updated.

pyp2rpm/bin.py
irushchyshyn: Added an extra check in f83a092

pyp2rpm/bin.py
irushchyshyn: I reworked the validation a bit, so this is not needed anymore.

mcyprian:
Could you please also update SYNOPSIS and AUTHOR sections of the man page? The rest of the updates look good to me :)

mcyprian (approved):

irushchyshyn:
Updated in eaeed79.
Sorry for that, did not notice that it was also changed.

PR #103, opened 2017-03-21
Include extras require in metadata extraction

mcyprian:
Including extras require listed in setup.py in build requires can prevent tests of doc building failures in some cases. PR also contains updated integration tests data.

### DIFF
--- a/command/extract_dist.py
+++ b/command/extract_dist.py
@@ -11,7 +11,7 @@ class extract_dist(Command):
     class_metadata = None
 
     def __init__(self, *args, **kwargs):
-        """Metadata dictionary is created, all the metadata attributes, that were 
+        """Metadata dictionary is created, all the metadata attributes, that were
         not found are set to default empty values. Checks of data types are performed.
         """
         Command.__init__(self, *args, **kwargs)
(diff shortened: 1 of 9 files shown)


irushchyshyn:
The changes look reasonable to me. Are there a lot of cases which fail due to missing extras_require?
Is it intended only for `SetupPyMetadataExtractor`? Should not it work the same way when extracting matadata from wheel archive?

command/extract_dist.py
```diff
@@ -22,6 +22,14 @@ def __init__(self, *args, **kwargs):
                      'packages', 'py_modules', 'scripts']:
             self.metadata[attr] = to_list(getattr(self.distribution, attr, []))
 
+        try:
+            self.metadata['extras_require'] = []
+            for v in getattr(self.distribution, 'extras_require', {}).values():
+                self.metadata['extras_require'] += to_list(v)
+        except AttributeError:
```
irushchyshyn: Should not we also catch `ValueError` here (which can be raised in `to_list`) or is it a rare case?

mcyprian: Right, I will add it.

mcyprian:
Maybe it is not obvious, but extras are already supported in wheel extractor. Format of metadata is a bit different in metadata.json file, "extra" is only a special tag of package included in dependencies.

mcyprian:
I added a trivial logic to distinguish between runtime and build extra requirements.

irushchyshyn:
Yeah, I did not see that it is already supported in wheel extractor.
Looks good!

PR #109, opened 2017-04-11
Summary processing

mcyprian:
- shorten summary if too long
- remove quotes and whitespaces
- fix unwanted removing of single dashes from description and summary

### DIFF
--- a/pyp2rpm/metadata_extractors.py
+++ b/pyp2rpm/metadata_extractors.py
@@ -28,6 +28,16 @@ except ImportError:
 
 logger = logging.getLogger(__name__)
 
+def cut_to_length(text, length, delim):
+    """Shorten given text on first delimiter after given number
+    of characters.
+    """
+    cut = text.find(delim, length)
+    if cut > -1:
+        return text[:cut]
+    else:
+        return text
+
 
 def pypi_metadata_extension(extraction_fce):
     """Extracts data from PyPI and merges them with data from extraction method."""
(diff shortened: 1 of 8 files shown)


hroncok (commented):
I'd make it so for summary, to cut it before first newline unconditionally, multiline summaries are bogus anyway.

pyp2rpm/metadata_extractors.py
```diff
@@ -28,6 +28,22 @@
...
+    """Shorten given text on first delimiter after given number
+    of characters.
+    """
+    if not isinstance(delims, list):
+        delims = [delims]
+
+    for delim in delims:
+        cut = text.find(delim, length)
+        if cut != -1:
+            break
```
hroncok: This does not cut it after first delimiter. See:

```python
>>> cut_to_length("foo bar. zap: foo", 6, ['.', ':'])
'foo bar'
>>> cut_to_length("foo bar. zap: foo", 6, [':', '.'])
'foo bar. zap'

```

mcyprian: Is the issue that docstring doesn't match the implementation or the implementation itself? Only one delimiter is always used as an argument at this moment.

hroncok: Well if the current behavior is desired, update the docstring.

mcyprian:
I changed the implementation of cut_to_length function to be as simple as possible for now.

mcyprian:
@hroncok can this be merged?

hroncok (approved):

hroncok:
Go ahead.

PR #110, opened 2017-04-11
Include license in doc subpackage

mcyprian:
- Add all license files to %{name}-doc subpackage
- Move sphinx documentation to doc subpackage in all the templates

### DIFF
--- a/pyp2rpm/templates/epel6.spec
+++ b/pyp2rpm/templates/epel6.spec
@@ -36,6 +36,12 @@ Summary:        {{ data.summary }}
 %description -n {{ data.pkg_name|macroed_pkg_name(data.srcname)|name_for_python_version(pv) }}
 {{ data.description|truncate(400)|wordwrap }}
 {%- endcall %}
+{%- if data.sphinx_dir %}
+%package -n {{ data.pkg_name|macroed_pkg_name(data.srcname)|name_for_python_version(pv, True) }}-doc
+Summary:        {{ data.name }} documentation
+%description -n {{ data.pkg_name|macroed_pkg_name(data.srcname)|name_for_python_version(pv, True) }}-doc
+Documentation for {{ data.name }}
+{%- endif %}
 
 %prep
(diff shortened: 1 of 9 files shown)


pyp2rpm/templates/epel6.spec
```diff
@@ -143,7 +143,13 @@ popd
 {{ '%{python2_sitelib}'|sitedir_for_python_version(pv) }}/{{ underscored_or_pypi(data.name, data.underscored_name) }}-%{version}-py?.?.egg-info
 {%- endif %}
 {% endcall %}
-
+{%- if data.sphinx_dir %}
+%files -n {{ data.pkg_name|macroed_pkg_name(data.srcname)|name_for_python_version(pv, True) }}-doc
+%doc html
+{%- if data.doc_license %}
+%license {{data.doc_license|join(' ')}}
```
hroncok, line 156: Is `%license` even available in epel6?

pyp2rpm/templates/epel7.spec
```diff
@@ -121,6 +121,9 @@ ln -sf %{_bindir}/{{ script|script_name_for_python_version(pv) }} %{buildroot}/%
 {%- if data.sphinx_dir %}
 %files -n python-%{pypi_name}-doc
 %doc html 
+{%- if data.doc_license %}
+%license {{data.doc_license|join(' ')}}
```
hroncok, line 125: Is `%license` even available in epel7?

pyp2rpm/templates/epel6.spec
mcyprian: Yes it is available in EPEL, [documentation](https://fedoraproject.org/wiki/EPEL:Packaging#The_.25license_tag) says it is mapped to %doc.  This macro is present in some of the specfiles for EPEL [python-pip](http://pkgs.fedoraproject.org/cgit/rpms/python-pip.git/tree/python-pip.spec?h=el6#n207) for example.

hroncok (approved):

PR #111, opened 2017-04-18
Fix spec2scl integration and require a new version

irushchyshyn:


mcyprian (approved):
Looks good, thank you

PR #127, opened 2017-07-04
Add name convertor based on automatic provides

mcyprian:
This convertor uses standardized name format of the dependencies based on virtual Provides created during RPM build: https://fedoraproject.org/wiki/Changes/Automatic_Provides_for_Python_RPM_Packages

There is a command line option to enable/disable this convertor, it is disabled by default at this moment. It can be used by default for Fedora after F27 mass rebuild.

[Here](https://copr.fedorainfracloud.org/coprs/mcyprian/pypi_provides_test/builds/) is the test build of spec file listing dependencies in the canonical format.

### DIFF
--- a/pyp2rpm.1
+++ b/pyp2rpm.1
@@ -54,6 +54,9 @@ Version of the package to download (ignored for local files).
 .B "\--venv / --no-venv \"
 Enable / disable metadata extraction from virtualenv.
 .TP
+.B "\--autonc/ --no-autonc\"
+Enable / disable using automatic provides with a standardized name in dependencies declaration.
+.TP
 \fB\-\-sclize\fR
 Convert tags and macro definitions to SCL\-style
 using `spec2scl` module. NOTE: SCL related options
--- a/pyp2rpm/name_convertor.py
+++ b/pyp2rpm/name_convertor.py
@@ -4,10 +4,9 @@ try:
     import dnf
 except ImportError:
     dnf = None
+from pkg_resources import safe_name
 
 from pyp2rpm import settings
-from pyp2rpm import utils
-from pyp2rpm.logger import LoggerWriter
 
 
 logger = logging.getLogger(__name__)
@@ -36,6 +35,7 @@ class NameConvertor(object):
             Versioned name or the original name if given version is None.
         """
         regexp = re.compile(r'^python(\d*|)-(.*)')
+        auto_provides_regexp = re.compile(r'^python(\d*|)dist(.*)')
 
         if not version or version == settings.DEFAULT_PYTHON_VERSION and not default_number:
             found = regexp.search(name)
@@ -50,6 +50,9 @@ class NameConvertor(object):
 
             if regexp.search(name):
                 versioned_name = re.sub(r'^python(\d*|)-', 'python{0}-'.format(version), name)
+            elif auto_provides_regexp.search(name):
+                versioned_name = re.sub(r'^python(\d*|)dist', 'python{0}dist'.format(version), name)
+
             else:
                 versioned_name = 'python{0}-{1}'.format(version, name)
             if epel and version != settings.DEFAULT_PYTHON_VERSION:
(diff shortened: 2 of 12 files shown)


pyp2rpm.1
```diff
@@ -54,6 +54,9 @@ Version of the package to download (ignored for local files).
 .B "\--venv / --no-venv \"
 Enable / disable metadata extraction from virtualenv.
 .TP
+.B "\--autonc/ --no-autonc\"
+Enable / disable name convertor using automatic provides.
```
irushchyshyn: Using `name convertor` in options description is a bit confusing for someone who for example does not know that we have `NameConvertor` class.
What about something like 'Enable / disable using automatic provides with a standardized name in dependencies declaration'? This is at least referenced in the [guidelines](https://fedoraproject.org/wiki/User:Torsava/PackagingDrafts/PythonProvides#Automatic_Provides_with_a_standardized_name).

pyp2rpm/templates/fedora.spec
```diff
@@ -33,9 +33,9 @@ Summary:        %{summary}
 {{ data.description|truncate(400)|wordwrap }}
 {% endfor -%}
 {%- if data.sphinx_dir %}
-%package -n {{ data.pkg_name|macroed_pkg_name(data.srcname)|name_for_python_version(pv, True) }}-doc
+%package -n {{ data.pkg_name|macroed_pkg_name(data.srcname)|name_for_python_version(None, True) }}-doc
 Summary:        {{ data.name }} documentation
-%description -n {{ data.pkg_name|macroed_pkg_name(data.srcname)|name_for_python_version(pv, True) }}-doc
+%description -n {{ data.pkg_name|macroed_pkg_name(data.srcname)|name_for_python_version(None .base_python_version, True) }}-doc
```
irushchyshyn: This seems wrong: `None .base_python_version` 
And why do we set the version to `None` here?

pyp2rpm/name_convertor.py
```diff
@@ -200,3 +202,13 @@ def rpm_name(self, name, python_version=None):
...
     return name.lower().replace('-', '').replace('_', '')
+
+
+class AutoProvidesNameConvertor(NameConvertor):
+    """Name convertor based on Automatic Provides for Python RPM Packages."""
+
+    non_alphanumeric_regexp = re.compile(r'[^0-9a-zA-Z]+')
+
+    def rpm_name(self, name, python_version=settings.DEFAULT_PYTHON_VERSION):
+        canonical_name = self.non_alphanumeric_regexp.sub('-', name).lower()
```
irushchyshyn: Please use setuptool's [safe_name](https://github.com/pypa/setuptools/blob/master/pkg_resources/__init__.py#L1376) function, which is the one [used](https://github.com/rpm-software-management/rpm/blob/master/scripts/pythondistdeps.py#L136) to add an automatic provide for the package (if you trace down `dist.key`).
You can import it as: `from pkg_resources import safe_name`
Basically the only difference is that the dot is also allowed in the canonical name. And for example `zope.testing` should not be converted to `zope-testing`. Please also add this example to `test_rpm_name` parameterization.

irushchyshyn (commented):
Can you please also update README page with the new option?

Apart from a couple of small comments, it looks good!

pyp2rpm/templates/fedora.spec
mcyprian: This part of template is out of for loop scope, pv is undefined so the version argument used to be an empty string. Passing None to the filter doesn't change the behavior of the template at all.  Only one doc subpackage is created for both versions in case the package includes sphinx documentation and the name of the package is simply python-%{pypi_name}-doc. I am not exactly sure if this violates the last update of Fedora Packaging Guidelines for Python but it is not part of the issue, this PR intends to fix.

irushchyshyn: Fair enough.

mcyprian: Maybe I should move this to the separate commit.

irushchyshyn:
Looks good! 👍

Issue #128, opened 2017-07-04
Stop using python- prefixed dependencies


mcyprian:
Generated spec files don't follow the latest [Packaging Guidelines for Python](https://fedoraproject.org/wiki/Packaging:Python) . Name convertors should always use versioned prefix (python2- or python3-) when converting names of dependencies.

PR #129, opened 2017-07-07
Dependencies with versioned prefix

mcyprian:
Convert names of dependencies using versioned prefix (python2-) to follow the latest Packaging Guidelines for Python. Fixes issue #128

### DIFF
--- a/tests/test_data/python-sphinx.spec
+++ b/tests/test_data/python-sphinx.spec
@@ -13,8 +13,8 @@ Source0:        https://files.pythonhosted.org/packages/source/S/%{pypi_name}/%{
 BuildArch:      noarch
  
 BuildRequires:  python2-devel
-BuildRequires:  python-setuptools
-BuildRequires:  python-sphinx
+BuildRequires:  python2-setuptools
+BuildRequires:  python2-sphinx
  
 BuildRequires:  python3-devel
 BuildRequires:  python3-setuptools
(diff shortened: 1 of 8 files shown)


irushchyshyn (approved):

tests/test_data/python-sphinx.spec
```diff
@@ -67,19 +67,19 @@ Requires:       python3-Jinja2 >= 2.3
...
+Requires:       python3-sqlalchemy >= 0.9
+Requires:       python3-whoosh >= 2.0
+Requires:       python3-colorama >= 0.3.5
 Requires:       python3-nose
 Requires:       python3-mock
 Requires:       python3-simplejson
 Requires:       python3-html5lib
-Requires:       python3-sqlalchemy >= 0.9
-Requires:       python3-whoosh >= 2.0
-Requires:       python3-colorama >= 0.3.5
```
irushchyshyn, line 82: This is not related to the fix in this pull request, but I am curious if the Requires and scripts (below) output should be deterministic. So that with every pyp2rpm run, we have the relevant changes in diff, without the same lines moving around.
@mcyprian what do you think?

hroncok: Sort it alphabetically?

irushchyshyn: As an option, yes.

mcyprian: @irushchyshyn your proposal does make sense to me.

Issue #126, opened 2017-06-29
Localized date in %changelog


hroncok:
```console
$ locale
LANG=cs_CZ.utf8
LC_CTYPE="cs_CZ.utf8"
LC_NUMERIC="cs_CZ.utf8"
LC_TIME="cs_CZ.utf8"
LC_COLLATE="cs_CZ.utf8"
LC_MONETARY="cs_CZ.utf8"
LC_MESSAGES="cs_CZ.utf8"
LC_PAPER="cs_CZ.utf8"
LC_NAME="cs_CZ.utf8"
LC_ADDRESS="cs_CZ.utf8"
LC_TELEPHONE="cs_CZ.utf8"
LC_MEASUREMENT="cs_CZ.utf8"
LC_IDENTIFICATION="cs_CZ.utf8"
LC_ALL=

$ ./mybin.py -b 3 blurb
...
%changelog
* Čt čen 29 2017 Miro Hrončok <mhroncok@redhat.com> - 1.0.0.post1-1
- Initial package.
```

This is on Fedora 26. i don't think it happened before on Fedora 25.

mcyprian:
I cannot reproduce this for some reason. I am running F26 and my locale output is exactly the same ^^.

hroncok:
You mean you cannot?

mcyprian:
Yes, of course.

hroncok:
Funny enough, I can reproduce, but running he code that produces this isolated, will not reproduce:

```pycon
>>> import time
>>> time.strftime('%a %b %d %Y', time.gmtime())
'Tue Jul 18 2017'
```

hroncok:
So for some reason, when I put this in front of that call:

```python
print(locale.getlocale(locale.LC_TIME))
```

It gives me `('cs_CZ', 'UTF-8')`. I don't know when this is being taken from, but it can be reset fairly simple, PR will follow.

PR #131, opened 2017-07-18
Use C LC_TIME locale when getting a datestamp

hroncok:
Fixes https://github.com/fedora-python/pyp2rpm/issues/126

### DIFF
--- a/pyp2rpm/utils.py
+++ b/pyp2rpm/utils.py
@@ -1,4 +1,6 @@
+import contextlib
 import functools
+import locale
 import logging
 import os
 import subprocess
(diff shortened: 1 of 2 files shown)


pyp2rpm/utils.py
```diff
@@ -200,3 +202,12 @@ def console_to_str(s):
...
         return s
+
+
+@contextlib.contextmanager
+def c_time_locale():
+    """Context manager with C LC_TIME locale"""
+    old_time_locale = locale.getlocale(locale.LC_TIME)
+    locale.setlocale(locale.LC_TIME, 'C')
+    yield
+    locale.setlocale(locale.LC_TIME, old_time_locale)
```
hroncok, line 213: This may be in try-finally block, but that might be an overkill.

mcyprian:
LGTM

PR #133, opened 2017-07-21
Fix README options respesentation

irushchyshyn:
This is a follow up PR for #127.

The README was updated with the newly added option, but I did not check the representation. So now [README](https://github.com/fedora-python/pyp2rpm/blob/master/README.md) looks completely wrong:
image


mcyprian:
Thanks ;)

PR #135, opened 2017-08-12
Use ordered lists to store requires, scripts, modules and packages

mcyprian:
This prevents order changes of metadata items between subsequent pyp2rpm runs. Requires, scripts, modules and packages are stored in list instead of set, and sorted.


pyp2rpm/metadata_extractors.py
```diff
@@ -300,7 +301,8 @@ def runtime_deps(self):  # install_requires
         if self.metadata['entry_points'] and 'setuptools' not in install_requires:
             install_requires.append('setuptools')  # entrypoints
 
-        return self.name_convert_deps_list(deps_from_pyp_format(install_requires, runtime=True))
+        return self.name_convert_deps_list(deps_from_pyp_format(
+            sorted(install_requires), runtime=True))
```
irushchyshyn: `install_requires` can also contain smth like `alabaster<0.8,>=0.7`, which after converting from pyp format transforms into 
```
Requires:       python3-alabaster < 0.8
Requires:       python3-alabaster >= 0.7
```
And as we sort before converting, those are still randomly switching places. Maybe we could do sorting after or while converting?

irushchyshyn (commented):
Currently in our [integration tests](https://github.com/fedora-python/pyp2rpm/blob/master/tests/test_integration.py#L38) we do specs comparison as set of lines. This is why the randomness is not identified in tests.
So you can change it back to lists comparison, to see how it works:
```python
assert res.stdout.split('\n')[1:-4] == self.spec_content.split('\n')[1:-4]
```

irushchyshyn:
Looks good 👍 Thanks!
Please squash and merge.

mcyprian:
Thanks for the review @irushchyshyn.

PR #136, opened 2017-09-05
Templates comment update

mcyprian:
- Update of comment in install section of templates
- Use %_topdir macro value to set default SAVE_DIR


irushchyshyn (approved):
Looks good! 👍

PR #139, opened 2017-09-12
Use output of rpm --eval command, in case import of rpm module fails

mcyprian:
Order of the methods to get DEFAULT_PKG_SAVE_PATH is following:
- rpm Python API (should work on systems where is yum/dnf installed)
- rpm --eval command
- hardcoded path ~/rpmbuild (nonrpm environments e.g. travis)

Some functions from utils had to be moved elsewhere to prevent circular dependency between utils and settings modules.

### DIFF
--- a/pyp2rpm/metadata_extractors.py
+++ b/pyp2rpm/metadata_extractors.py
@@ -58,7 +58,7 @@ def license_from_trove(trove):
     """
     license = []
     for classifier in trove:
-        if 'License' in classifier != -1:
+        if 'License' in classifier:
             stripped = classifier.strip()
             # if taken from EGG-INFO, begins with Classifier:
             stripped = stripped[stripped.find('License'):]
--- a/pyp2rpm/metadata_extractors.py
+++ b/pyp2rpm/metadata_extractors.py
@@ -40,6 +40,50 @@ def cut_to_length(text, length, delim):
         return text
 
 
+def get_interpreter_path(version=None):
+    """Return the executable of a specified or current version."""
+    if version and version != str(sys.version_info[0]):
+        return settings.PYTHON_INTERPRETER + version
+    else:
+        return sys.executable
+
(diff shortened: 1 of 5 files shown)


pyp2rpm/metadata_extractors.py
```diff
@@ -40,6 +40,50 @@ def cut_to_length(text, length, delim):
...
+    """Finds out license from list of trove classifiers.
+    Args:
+        trove: list of trove classifiers
+    Returns:
+        Fedora name of the package license or empty string, if no licensing
+        information is found in trove classifiers.
+    """
+    license = []
+    for classifier in trove:
+        if 'License' in classifier != -1:
```
irushchyshyn: I believe `!= -1` part is redundant?

pyp2rpm/utils.py
```diff
@@ -211,3 +134,25 @@ def c_time_locale():
...
+
+
+def rpm_eval(macro):
+    """Get value of given macro using rpm tool"""
+    try:
+        value = subprocess.Popen(
+            ['rpm', '--eval', macro],
+            stdout=subprocess.PIPE).communicate()[0].strip()
+    except OSError:
+        logger.error('Failed to get value of {0} rpm macro'.format(
```
irushchyshyn, line 146: I think this should be *warning* for the user, also saying that the default path will be used.

irushchyshyn (commented):
Looks good.
I believe if those functions are used only in `metadata_extractors.py` then they should not have been in utils in the first place, so moving them there is fine and fixes circular dependency issue. But we'll never be able to use settings in utils, oh well:)

How about adding some tests for `get_default_save_path` and `rpm_eval`?

irushchyshyn (approved):
Thanks for the fixes. Looks good.
Please squash the fixup and merge at will.

PR #142, opened 2017-10-12
Python 3 as default python version for Fedora

mcyprian:
- Each disto (template) has its own default python versions
- Python 3 is now default for Fedora
- Default executables policy change


pyp2rpm/bin.py
```diff
@@ -80,15 +80,17 @@ def get_scl_help_record(self, ctx):
...
-              help='Base Python version to package for (default: "{0}").'.format(
+              help='Base Python version to package for (fedora default: "{0}").'.format(
                   settings.DEFAULT_PYTHON_VERSION),
               default=None,
               metavar='BASE_PYTHON')
 @click.option('-p',
-              help='Additional Python versions to include in the specfile (e.g -p3 for %{0}).'
-              'Can be specified multiple times (default: "{1}"). Specify additional version '
+              help='Additional Python versions to include in the specfile '
+              '(e.g -p2 for {0}). Can be specified multiple times '
```
irushchyshyn: `{0}` formatting is not needed here, just do `(e.g -p2 for python2 subpackage).`

tests/test_data/python-sphinx.spec
```diff
@@ -103,63 +103,29 @@ rm -rf %{pypi_name}.egg-info
...
-cp %{buildroot}/%{_bindir}/sphinx-apidoc %{buildroot}/%{_bindir}/sphinx-apidoc-%{python3_version}
-ln -s %{_bindir}/sphinx-apidoc-%{python3_version} %{buildroot}/%{_bindir}/sphinx-apidoc-3
-cp %{buildroot}/%{_bindir}/sphinx-autogen %{buildroot}/%{_bindir}/sphinx-autogen-%{python3_version}
-ln -s %{_bindir}/sphinx-autogen-%{python3_version} %{buildroot}/%{_bindir}/sphinx-autogen-3
-cp %{buildroot}/%{_bindir}/sphinx-build %{buildroot}/%{_bindir}/sphinx-build-%{python3_version}
-ln -s %{_bindir}/sphinx-build-%{python3_version} %{buildroot}/%{_bindir}/sphinx-build-3
-cp %{buildroot}/%{_bindir}/sphinx-quickstart %{buildroot}/%{_bindir}/sphinx-quickstart-%{python3_version}
-ln -s %{_bindir}/sphinx-quickstart-%{python3_version} %{buildroot}/%{_bindir}/sphinx-quickstart-3
-
 %py2_install
```
irushchyshyn, line 113: I really think that we should do `rm -rf %{buildroot}%{_bindir}/*` after Py2 install in such cases.

I have already seen a number of cases when the scripts did not get overwritten, resulting in a Py3 subpackage having a dependency on Py2. See the Note in the [Python RPM porting guide](https://python-rpm-porting.readthedocs.io/en/latest/application-modules.html#install). 

We could add smth like this after install:
```
{%- if data.scripts and pv != data.base_python_version %}
rm -rf %{buildroot}%{_bindir}/*
{%- endif %}
```

pyp2rpm/templates/fedora.spec
```diff
@@ -47,76 +46,70 @@ rm -rf %{pypi_name}.egg-info
...
 {%- if pv == data.base_python_version %}
+{%- for script in data.scripts %}
 %{_bindir}/{{ script }}
-{%- endif %}
-%{_bindir}/{{ script|script_name_for_python_version(pv) }}
-%{_bindir}/{{ script|script_name_for_python_version(pv, True) }}
 {%- endfor %}
+{%- endif %}
 {%- if data.py_modules %}
 {% for module in data.py_modules -%}
```
irushchyshyn: Please use here `{%- ...` for new line control.

Otherwise it looks like this (example from Django):
```
%files -n python3-%{pypi_name}
%license LICENSE.txt
%doc DESCRIPTION.rst
%{_bindir}/django-admin.py
%{_bindir}/django-admin

%{python3_sitelib}/__pycache__/*
%{python3_sitelib}/django.py
%{python3_sitelib}/django
%{python3_sitelib}/%{pypi_name}-%{version}-py?.?.egg-info
```

pyp2rpm.1
```diff
@@ -28,10 +28,10 @@ Search order is: 1) filesystem, 2) default templates.
 Default distro whose conversion rules to use (default: "fedora"). Default templates have their rules associated and ignore this.
 .TP
 .B "\-b \-\-BASE_PYTHON"
-Base Python version to package for (default: "2").
+Base Python version to package for (fedora default: "3").
 .TP
 .B "\-p \-\-PYTHON_VERSIONS"
-Additional Python versions to include in the specfile (e.g. -p3 f or %{?with_python3}). Specify additional version or use -b explicitly to disable default.
+Additional Python versions to include in the specfile (e.g. -p2 for python2 subpackage}). Specify additional version or use -b explicitly to disable default.
```
irushchyshyn: Please remove the closing `}`.

pyp2rpm/bin.py
```diff
@@ -159,8 +161,12 @@ def main(package, v, d, s, r, proxy, srpm, p, b, o, t, venv, autonc, sclize, **s
         register_console_log_handler()
 
     distro = o
-    if t in settings.KNOWN_DISTROS:
+    if t and os.path.splitext(t)[0] in settings.KNOWN_DISTROS:
         distro = t
+    elif t and not b or p:
```
irushchyshyn: When I do `./mybin.py -p2 Sphinx` I get:
```
Usage: mybin.py [OPTIONS] PACKAGE

Error: Default python versions for template None are missing in settings, add them or use flags -b/-p to set python versions.
```

This line should probably be changed to `elif t and not (b or p):`

irushchyshyn (commented):
That probably took a lot of work 👍 

I have added a couple of small comments and please take a look at failing travis build (the tests pass in py27 env locally).

mcyprian:
Travis integration test failed because metadata extraction using alternative python3 interpreter (/usr/bin/python3) didn't work in Ubuntu Trusty container used by Travis. I guess it will be better to run Travis tests inside custom Fedora container. This will make Travis builds easily reproducible in local environment in similar cases and we will be able to test rpm and dnf specific features.

irushchyshyn (approved):
Looks good!👍  Please squash the fixups and merge at will!

I would suggest though in future to make changes for separate issues in separate PRs so they are not too big.

Issue #145, opened 2017-10-23
setup.py not found when trying pyp2rpm on jrti-game


encukou:
I tried pyp2rpm on my package, `jrti_game`, without success:

```
$ pyp2rpm jrti_game
Package "jrti_game" could not be found on PyPI.
$ pyp2rpm jrti-game
setup.py not found, maybe local_file is not proper source archive.
```

I believe it may have to do with package name normalization, see https://github.com/fedora-python/pyp2rpm/issues/144

(Also, what's `local_file`?)

irushchyshyn:
I believe the following case was fixed in #98 and is available starting version 3.2.2.

> pyp2rpm jrti-game
> setup.py not found, maybe local_file is not proper source archive.

Which version of pyp2rpm was used?

encukou:
I see! I use the version in Fedora 26.
```
pyp2rpm-3.2.1-3.fc26.noarch
```

It works with the pip-installed one. Sorry for the noise!

irushchyshyn:
> (Also, what's local_file?)

Good question:)

PR #146, opened 2017-10-23
Fix error message when setup.py cannot be found

irushchyshyn:
Reference `self.local_file` in an error message when setup.py can't be found. E.g change form:

```
 $ ./mybin.py jrti-game
Failed to synchronize cache for repo 'fedora-skype', disabling.
setup.py not found, maybe local_file is not proper source archive.
```
to:
```
 $ ./mybin.py jrti-game
Failed to synchronize cache for repo 'fedora-skype', disabling.
setup.py not found, maybe /home/ishcherb/rpmbuild/SOURCES/jrti_game-1.0.tar.gz is not proper source archive.
```
Also reported in #145.


mcyprian:
LGTM.

PR #147, opened 2017-11-02
Set max line length to 80 characters

mcyprian:
Change max line from 100 to 80 characters to be able to use automated code linters without special configuration. 
- Fix various small flake8 errors
- Remove outdated TODO file.


pyp2rpm/archive.py
```diff
@@ -134,12 +136,14 @@ def extractor_cls(self):
         elif self.is_zip:
             file_cls = ZipFile
         else:
-            logger.info("Couldn't recognize archive suffix: {0}.".format(self.suffix))
+            logger.info("Couldn't recognize archive suffix: {0}.".format(
+                    self.suffix))
```
irushchyshyn: The indentation is wrong here.

pyp2rpm/archive.py
```diff
@@ -153,34 +157,43 @@ def get_content_of_file(self, name, full_path=False):  # TODO: log if file can't
...
+    # TODO: maybe implement this using get_files_re
+    def has_file_with_suffix(self, suffixes):
         """Finds out if there is a file with one of suffixes in the archive.
         Args:
             suffixes: list of suffixes or single suffix to look for
         Returns:
-            True if there is at least one file with at least one given suffix in the archive,
+            True if there is at least one file with at least one given suffix
+            in the archive,
             False otherwise (or archive can't be opened)
```
irushchyshyn: Please move this one line up, so it looks like a continuation of the above :)

pyp2rpm/bin.py
```diff
@@ -74,32 +80,33 @@ def get_scl_help_record(self, ctx):
...
               default=None,
               metavar='BASE_PYTHON')
 @click.option('-p',
               help='Additional Python versions to include in the specfile '
-              '(e.g -p2 for python2 subpackage). Can be specified multiple times '
-              '(fedora default: "{0}"). Specify additional version '
-              'or use -b explicitly to disable default.'.format(
-                  settings.DEFAULT_PYTHON_VERSIONS[
-                      settings.DEFAULT_DISTRO][1]),
+                   '(e.g -p2 for python2 subpackage). Can be specified '
```
irushchyshyn: The indentation for this help message is bigger from the others, why so?
Same for  `--list-file` option.

pyp2rpm/bin.py
```diff
@@ -110,51 +117,68 @@ def get_scl_help_record(self, ctx):
...
-def main(package, v, d, s, r, proxy, srpm, p, b, o, t, venv, autonc, sclize, **scl_kwargs):
+def main(package, v, d, s, r, proxy, srpm, p, b, o, t, venv, autonc, sclize,
+         **scl_kwargs):
     """Convert PyPI package to RPM specfile or SRPM.
 
     \b
     \b\bArguments:
-    PACKAGE             Provide PyPI name of the package or path to compressed source file."""
+    PACKAGE             Provide PyPI name of the package or path to compressed
+    source file.
```
irushchyshyn: This looks weird in the help message. Maybe change so it looks:
image

pyp2rpm/convertor.py
```diff
@@ -76,19 +79,21 @@ def template_py_vers(self):
...
-            if any(int(ver[0]) > 2 for ver in self.python_versions + ([
-                    self.base_python_version] if self.base_python_version else [])):
+            requested_versions = self.python_versions
+            if self.base_python_version:
+                requested_versions += [self.base_python_version]
+            if any(int(ver[0]) > 2 for ver in requested_versions):
                 sys.stderr.write(
-                    "Invalid version, major number of python version for EPEL6 "
-                    "spec file must not be greater than 2.\n")
+                    "Invalid version, major number of python version for"
```
irushchyshyn: Please keep a space in the end on the line.

pyp2rpm/convertor.py
```diff
@@ -172,7 +180,9 @@ def getter(self):
                     self.package,
                     self.save_dir)
             else:
-                logger.debug('{0} doesnt exists as local file trying PyPI.'.format(self.package))
+                logger.debug(
+                    '{0} doesnt exists as local file trying PyPI.'.format(
```
irushchyshyn: Can you also fix `doesnt exists` ->`does not exist`?

pyp2rpm/package_getters.py
```diff
@@ -128,21 +135,22 @@ def __init__(self, client, name, version=None, save_dir=None):
...
 
         # if version is specified, will check if such version exists
         if version and self.client.release_urls(name, version) == []:
             raise exceptions.NoSuchPackageException(
-                'Package with name "{0}" and version "{1}" could not be found on PyPI.'.format(
-                    name, version))
-            logger.error(
-                'Package with name "{0}" and version "{1}" could not be found on PyPI.'.format(
-                    name, version))
+                'Package with name "{0}" and version "{1}" could not be '
```
irushchyshyn, line 148: Maybe move it to a variable and use in both cases?

irushchyshyn (commented):
Why is this needed? Are we going to introduce some style checker in builds or smth?

It looks good, I have added a bunch of small nitpick comments though :)

pyp2rpm/package_getters.py
mcyprian: I've realized this message won't ever be written into log, because an exception is raised one line before :) Exception appears in the log so I think it is not necessary to duplicate the message.

mcyprian:
Style check in each build might be a good idea, but my primary goal was to make code more readable, and speed up our work flows by adding an possibility to simply use autopep8 or similar code  formatter after each change in the code without any configuration.

pyp2rpm/package_getters.py
irushchyshyn: Oh, right :)

irushchyshyn (approved):
Thanks for the fixes! Looks good 👍 

Please rebase and merge.

mcyprian:
Thanks :)

PR #149, opened 2017-11-14
Skip virtualenv tests if virtualenv-api is not installed

mcyprian:
virtualenv-api is an optional dependency of pyp2rpm. If it is not installed related tests will be skipped. This PR also contains small change in integration tests that prevents blocking of test_output directory. Integration tests can be run in parallel inside multiple virtual environments.

### DIFF
--- a/tests/test_data/python-StructArray_dnfnc.spec
+++ b/tests/test_data/python-StructArray_dnfnc.spec
@@ -47,9 +47,9 @@ rm -rf %{pypi_name}.egg-info
 
 %files -n python2-%{pypi_name}
 %doc 
-%{python2_sitearch}/structarray.so
+%{python2_sitearch}/%{pypi_name}
 %{python2_sitearch}/%{pypi_name}-%{version}-py?.?.egg-info
 
 %changelog
-* Mon Oct 23 2017 Michal Cyprian <mcyprian@redhat.com> - 0.1-1
+* Tue Nov 14 2017 Michal Cyprian <mcyprian@redhat.com> - 0.1-1
 - Initial package.
--- a/tests/test_virtualenv.py
+++ b/tests/test_virtualenv.py
@@ -2,10 +2,18 @@ import pytest
 import shutil
 import tempfile
 from flexmock import flexmock
-from pyp2rpm.virtualenv import (DirsContent, VirtualEnv, site_packages_filter,
-                                scripts_filter)
+try:
+    from pyp2rpm.virtualenv import (DirsContent,
+                                    VirtualEnv,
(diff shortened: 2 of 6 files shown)


tests/test_virtualenv.py
```diff
@@ -2,11 +2,19 @@
...
+    from pyp2rpm.virtualenv import (DirsContent,
+                                    VirtualEnv,
+                                    site_packages_filter,
+                                    scripts_filter)
+except ImportError:
+    VirtualEnv = None
 from pyp2rpm.name_convertor import NameConvertor
 from pyp2rpm.settings import DEFAULT_DISTRO, DEFAULT_PYTHON_VERSION
 
+pytestmark = pytest.mark.skipif(VirtualEnv is None,
```
hroncok, line 15: Shouldn't you actually use the marker? Or how does it work, what tests are skipped?

hroncok:
Have a look at https://docs.pytest.org/en/latest/tmpdir.html - maybe using pytest fixtures here would work better?

tests/test_virtualenv.py
mcyprian: The whole module is skipped as documented here https://docs.pytest.org/en/latest/skipping.html#skip-all-test-functions-of-a-class-or-module.

mcyprian:
I was looking in to tmdir fixture and my impression is that is is designed to be used in one particular method/function. I haven't found the way to use it in setup_method where TestFileEnvironment is created. At this moment we have only one test method in each class so it is also an possibility to move env creation to methods and use the fixture.

tests/test_data/python-StructArray_dnfnc.spec
```diff
@@ -47,9 +47,9 @@ rm -rf %{pypi_name}.egg-info
 
 %files -n python2-%{pypi_name}
 %doc 
-%{python2_sitearch}/structarray.so
+%{python2_sitearch}/%{pypi_name}
```
hroncok, line 50: I'm unsure how is this relevant for the change. Maybe I'm missing something?

mcyprian: [An integration test for StructArray](https://github.com/fedora-python/pyp2rpm/pull/149/commits/23bd40ac6bef5856516afe68e245f13c4a5d19c4#diff-7eb47fc18d4a6196f0940784484a5bf3R32) is now running with --no-venv switch. This was done in order to keep results of integration tests consistent in case virtualenv-api is not installed.

hroncok (approved):
Ok, all makes sense to me now.

mcyprian:
Thanks.

PR #150, opened 2017-11-21
BuildRequires enhancement when test suite is present

mcyprian:
Detection of test suite based on data from setup.py is not reliable. Check if the archive contains tests/ directory was added. In case the test suite is detected, all the install requires are added to BuildRequires as well to make sure all the dependencies to run integration tests will be available in the Buildroot.

### DIFF
--- a/pyp2rpm/metadata_extractors.py
+++ b/pyp2rpm/metadata_extractors.py
@@ -382,15 +382,20 @@ class SetupPyMetadataExtractor(LocalMetadataExtractor):
             install_requires, runtime=True)))
 
     @property
-    def build_deps(self):  # setup_requires + tests_require
-        """Same as runtime_deps, but build dependencies. Test requires
-        are included only if package contains test suite.
+    def build_deps(self):  # setup_requires [tests_require, install_requires]
+        """Same as runtime_deps, but build dependencies. Test and install
+        requires are included if package contains test suite to prevent
+        %check phase crashes because of missing dependencies
 
         Returns:
             list of build dependencies of the package
         """
-        build_requires = self.metadata['setup_requires'] + self.metadata[
-            'tests_require']
+        build_requires = self.metadata['setup_requires']
+
+        if self.has_test_suite:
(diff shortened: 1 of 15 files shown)


pyp2rpm/metadata_extractors.py
```diff
@@ -584,7 +590,10 @@ def license(self):
 
     @property
     def has_test_suite(self):
-        return self.json_metadata.get('test_requires', False) is not False
+        test_dir = "{0}/tests".format(self.archive.top_directory, 'tests') in \
```
irushchyshyn: Can we move `test_dir` to be the property/method of `LocalMetadataExtractor` and reuse it for both cases?

Also if `test_dir` is a bool, it would be better to name it smth like `has_test_dir`.
Also the `format` method gets an extra useless argument `tests`.

pyp2rpm/metadata_extractors.py
```diff
@@ -371,14 +371,18 @@ def runtime_deps(self):  # install_requires
...
+        crashes because of missing dependencies
 
         Returns:
             list of build dependencies of the package
         """
         build_requires = self.metadata['setup_requires'] + self.metadata[
             'tests_require']
+
+        if self.has_test_suite:
+            build_requires += self.metadata['install_requires']
```
irushchyshyn: Can we also move adding `self.metadata['tests_require']` here from the line above?
I will not change the behavior, but it would look more logical.

irushchyshyn: Also why don't we add the same handling for `build_deps` in `WheelMetadataExtractor` to add `run_requires` if the test suite is present?

mcyprian:
Test suite detection was reimplemented to search for files with test_ prefix or _test suffix instead of tests directory. pytest does the similar thing when collecting test files. All the requested changes should be covered by two fixup commits.

irushchyshyn (approved):
Looks good 👍 
Please rebase and merge at will.

PR #154, opened 2017-12-06
Autonc as default

mcyprian:
I think we should make name convertor based on Automatic Provides the default for Fedora.


pyp2rpm/name_convertor.py
```diff
@@ -80,7 +80,7 @@ def rpm_versioned_name(cls, name, version, default_number=False):
                     version), '%{{python{0}_pkgversion}}'.format(version))
         return versioned_name
 
-    def rpm_name(self, name, python_version=None):
+    def rpm_name(self, name, python_version=None, pkg_name=False):
         """Returns name of the package converted to (possibly) correct package
            name according to Packaging Guidelines.
         Args:
```
irushchyshyn, line 86: Nitpick: please document `pkg_name` argument.

irushchyshyn (commented):
Looks good, however I find it strange that for Fedora `pyp2rpm <package> --autonc ` and `pyp2rpm <package>` does the same. 
What is the use of the `--autonc` then?

mcyprian:
Documentation of pkg_name argument added. Flag --autonc has no effect only in case fedora template is used. Someone might use other (custom) template with automatic provides based nc. Also in case someone uses --autonc at this moment removing it will brake his scripts. In my opinion we should keep it as it is for now.

irushchyshyn:
Looks good!👍 
Please do the rebase and merge.

irushchyshyn (approved):

mcyprian:
Thanks ;)

PR #155, opened 2017-12-07
configure PYTHONPATH when building docs with sphinx

TomasTomecek:
I'm assuming that tests will break; will address those issues once CI finishes

### DIFF
--- a/pyp2rpm/templates/epel6.spec
+++ b/pyp2rpm/templates/epel6.spec
@@ -57,7 +57,7 @@ find python{{pv}} -name '*.py' | xargs sed -i '1s|^#!python|#!%{__python{{pv}}}|
 {%- if data.sphinx_dir %}
 # generate html docs {# TODO: generate properly for other versions (pushd/popd into their dirs...)
 # #}
-{{ "sphinx-build"|script_name_for_python_version(data.base_python_version, False, False) }} {{ data.sphinx_dir }} html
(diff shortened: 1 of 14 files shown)


TomasTomecek:
Honestly, I have no idea why the tests fail, since I changed the spec files in `test_data`.

mcyprian:
Spec file examples using DNF based name_convertor have to be updated as well:
- python-Jinja2_base_dnfnc.spec
- python-Jinja2_dnfnc.spec
- python-Jinja2_epel6_dnfnc.spec
- python-Jinja2_epel7_dnfnc.spec
- python-sphinx_dnfnc.spec) 


In `python-Jinja2_autonc.spec` there is `doc` vs` docs` difference:
`
At index 73 diff: u'PYTHONPATH=${PWD} sphinx-build-3 docs html' != 'PYTHONPATH=${PWD} sphinx-build-3 doc html'`

You can use [commands from .travis.yml ](https://github.com/fedora-python/pyp2rpm/blob/master/.travis.yml#L6) to run test suite locally.

TomasTomecek:
Thanks Michal, good catch.

Should be okay now.

mcyprian:
Looks good to me, thank you.

TomasTomecek:
Thanks! I hope this is gonna be useful for more people and not just me.

hroncok:
BTW https://github.com/fedora-modularity/conu/pull/99

PR #156, opened 2017-12-07
Use the warehouse installation URL for XMLRPC client for PyPI

### LINKED ISSUE #144: Package names are not normalized
`pip` normalizes package names as per [PEP 503](https://www.python.org/dev/peps/pep-0503/#normalized-names), so `pip install flask` and `pip install Flask` will both work.
However, pyp2rpm doesn't seem to do this:

```console
$ pyp2rpm flask
Package "flask" could not be found on PyPI.
```

irushchyshyn:
Fixes #144 (package name normalization)


mcyprian:
LGTM.

PR #158, opened 2017-12-13
Fix pulling the lowest release version of the package from PyPI

irushchyshyn:
After we switched to https://pypi.org/pypi, the order in which `package_releases` are returned has changed, and we now get releases starting from oldest to the most recent. As we depend on that ordering, pyp2rpm was creating a spec file for the lowest version of the package.

The integration tests did not catch that, because we specify version with `-v` option to avoid broken tests every time the package gets updated on PyPI.


mcyprian:
Thanks for fixing this @irushchyshyn.

PR #165, opened 2018-03-17
Rebase Mageia template based on latest Python Packaging Policy

Conan-Kudo:
Mageia's packaging, as of Mageia 7, is similar to how Fedora packages Python modules.

With Mageia 7, we are also using the new dependency generator by default and Python 3 is now the default.


irushchyshyn (approved):
Looks good!
Do you need a new release for this?

Conan-Kudo:
@irushchyshyn It'd be appreciated, yes.

irushchyshyn:
@Conan-Kudo no problem, https://pypi.python.org/pypi/pyp2rpm/3.3.1.

Issue #163, opened 2018-01-25
For Fedora, consider making Python 3 only the default


hroncok:
Currently, when pyp2rpm is used without any switches, it creates a combined py3+py2 package. I was wondering whether we are at a point where python2 subpackage would only be created when explicitly asked for (or when pyp2rpm determines the package is py2 only).

I think there are packagers that simply include the python2 subpackage because it was created automatically, even if they don't actually need one. Those who need it would simply use the appropriate switch (`-p2`).

Thoughts?

Conan-Kudo:
This would be good for Fedora and Mageia, as we're considering the same change in Mageia, too.

However, I think this would require some coordination to figure out what to do about unversioned Python package names.

irushchyshyn:
This has been fixed for Fedora and Mageia. And I will do another release for it.
Closing, please reopen if needed.

PR #166, opened 2018-03-20
Python 3 the only default version for Fedora and Mageia

irushchyshyn:
#163

When used without any switches, pyp2rpm should create a spec file which builds only Python 3 subpackage (unless the Trove classifiers suggest it is Python 2 only).

To also build a Python 2 subpackage `-p2` option should be used.


mcyprian (commented):
Version selection is correct and logical for both Fedora and Mageia. The merge_versions() method is now also slightly more readable, good job. Please update the documentation of -p option in README.md file as well. When this is fixed, feel free to merge this.

irushchyshyn:
Thanks for the review @mcyprian !

PR #170, opened 2018-06-05
Add possibility to disable doc build for Fedora by external macro

dsilakov:
Sometimes it is convenient to disable generation of documentation for python modules. A common practice is to use "doc_guard" for this - enclose doc-related parts with `%if 0%{?with_doc}` condition (see e.g., https://github.com/redhat-openstack/easyfix/issues/6).

Default value for with_doc macro is set to 1; this patch also takes into account global _without_doc macro, so one can turn documentation on/off by changing a single macro in a build environment.


hroncok:
1. I think this creates unnecessary clutter in spec and should be optional
 2. Please use [bcond](http://rpm.org/user_doc/conditional_builds.html) based conditions

pyp2rpm/templates/fedora.spec
```diff
@@ -5,7 +5,11 @@
 %global srcname {{ data.srcname }}
 {%- endif %}
 {%- if data.sphinx_dir %}
-%global with_doc %{!?_without_doc:1}%{?_without_doc:0}
+%if 0%{?_without_doc}
+%bcond_with doc
+%else
+%bcond_without doc
```
hroncok, line 8: just keep this line; the if/else part is not needed at all.

hroncok:
Now should it be doc or docs?

dsilakov:
Not sure if Fedora has formalized policy on this. I kept in mind OpenStack specs from RDO, many of them use "doc" and none uses "docs"

patsevanton:
Hello!

Do you accept a pull request that adds a "- without_doc" file that disables the documentation build?

This is just an option. By default, pyp2rpm will also compile with documentation.

Issue #24, opened 2015-11-16
pyp2rpm doesnt seem to work with HTTP_PROXY env variable


laapsaap:
pyp2rpm doesnt seem to work with HTTP_PROXY env variable

```
$ export HTTP_PROXY=http://proxy:3128
$ export HTTPS_PROXY=https://proxy:3128
$ pyp2rpm elasticsearch
...
socket.error: [Errno 110] Connection timed out
```

laapsaap:
Nevermind I see there's a --proxy arg. :+1: 

But I do think it should with HTTP_PROXY env variable,  but that is just luxury :)

laapsaap:
I found the --proxy argument, but it isnt working. Stracing the process doesnt show me pyp2rpm trying to connect to a proxy.

gss2002:
Here is what I did to make this work behind our corporate proxies... 
[convert_py_proxy.txt](https://github.com/fedora-python/pyp2rpm/files/2275217/convert_py_proxy.txt)

PR #173, opened 2018-08-09
Fix pyp2rpm to respect --proxy or env http_proxy/https_proxy…

gss2002:
pyp2rpm doesnt seem to work with HTTP_PROXY or http_proxy or https_proxy or HTTPS_PROXY env variable and also doesn't work with --proxy. So after debugging and testing against a corporate enterprise proxy server I have come up with the following fix.

This fixes the following issue: fedora-python/pyp2rpm#24

### DIFF
--- a/pyp2rpm/convertor.py
+++ b/pyp2rpm/convertor.py
@@ -56,7 +56,37 @@ class Convertor(object):
         if not self.template.endswith('.spec'):
             self.template = '{0}.spec'.format(self.template)
         self.rpm_name = rpm_name
-        self.proxy = proxy
+        #check and set proxy settings
+
+        self.use_proxy = None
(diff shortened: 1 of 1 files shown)


gss2002:
Failure not from this change

pyp2rpm/convertor.py
```diff
@@ -56,7 +56,37 @@ def __init__(self, package=None, version=None,
         if not self.template.endswith('.spec'):
             self.template = '{0}.spec'.format(self.template)
         self.rpm_name = rpm_name
-        self.proxy = proxy
+        #check and set proxy settings
```
hroncok, line 59: Could this be a function/method on it's own?

gss2002: @hroncok I will look at doing that today. I was away on vaca sorry for slow response.

PR #177, opened 2019-01-09
Add full url to the pypi source for Source0

kekePower:
This PR adds the full URL to the pypi source for the Mageia spec.

I've tested it locally and it works and makes my job easier.


hroncok:
Wouldn't it be easier to adapt [%pypi_source](https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_source_files_from_pypi) ([source](https://src.fedoraproject.org/rpms/python-rpm-macros/blob/master/f/macros.python-srpm#_83)) on Mageia distro level?

kekePower:
@hroncok 
Thanks for the suggestion.

I've asked on our dev mailinglist if this is possible.

Cheers.

kekePower:
pypi_source has been added to our tools. Thanks for the suggestion.

Issue #180, opened 2019-03-21
Failed to extract data from paperwork-backend


QuLogic:
Trying to run pyp2rpm on [paperwork-backend](https://pypi.org/project/paperwork-backend/) fails:
```bash
$ pyp2rpm paperwork-backend
Failed to extract data from setup.py script.
```
It also fails in the same way running `mybin.py` from `master`.

QuLogic:
Or with `-s`:
```
$ pyp2rpm -s paperwork-backend
INFO  Pyp2rpm initialized.
INFO  Using ~/rpmbuild/SOURCES as directory to save source.
INFO  Downloaded package from PyPI: ~/rpmbuild/SOURCES/paperwork-backend-1.2.4.tar.gz.
INFO  Getting metadata from setup.py using SetupPyMetadataExtractor.
INFO  Running extract_dist command with: /usr/bin/python3
ERROR  Could not extract metadata with: /usr/bin/python3
INFO  Running extract_dist command with: /usr/bin/python2
ERROR  Could not extract metadata with: /usr/bin/python2
Failed to extract data from setup.py script.
```

hroncok:
The package has a setup.py that prints stuff. We inject a custom setup.py command that should print json and we try to parse the json. However this is not valid json:

```
extracted json data:
[...]

PR #181, opened 2019-03-21
Add a workaround for setup.py that prints

hroncok:
Fixes https://github.com/fedora-python/pyp2rpm/issues/180

Hopefully the tests will not break, I'm trying now locally.

If it passes, I'll add Python 3.7 to the test docker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,8 +8,7 @@ LABEL summary="Image for running automatic tests of pyp2rpm in Travis CI" \
       name="pyp2rpm-tests" \
       maintainer="Michal Cyprian <mcyprian@redhat.com>"
 
-RUN INSTALL_PKGS="gcc tox python34 python35 python2-dnf python2-virtualenv \
-                  python2-virtualenv-api python3-virtualenv python3-virtualenv-api" && \
+RUN INSTALL_PKGS="gcc tox python27 python34 python35 python36 python37" && \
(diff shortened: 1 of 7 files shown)


hroncok:
@QuLogic would you be able to review this? Especially if it helps you, but looking at the code would be nice as well.

Note: Ideally I would add a test for failed metadata_extractor report, to check it indeed reports the failed json to load, but given the state of this project, I'd rather not. I've added an integration test for paperwork-backend.

hroncok:
The integration test failures on 3.7 are unrelated, but related to Sphinx. I don't have the strength to solve them, maybe just skip them?

QuLogic:
This works for me on paperwork-backend and paperwork. I don't see anything wrong except some "misuse" of the logging API, but that's an existing problem all over the code base.

hroncok:
@encukou there are unrelated Sphinx package names test failures.

 * we can skip them and never get back to them again
 * or merge as is and probably never get beck to them either, but at least there will be no false sense of safety

While broken tests on CI can seriously hurt a project, the tests were broken here for months, having this on life support I rather be reminded about the failing tests once a year when we change stuff rather than no knowing about it.

(Also, actually investigating and fixing the failures is out of my possibilities now and given the state of the project, I think that's OK.)

Thanks.

encukou (approved):
That makes sense. I'm also unlikely to spend any significant time on pyp2rpm.

PR #187, opened 2019-07-07
Use logging.warning instead of the obsolete logging.warn.

gordonmessmer:
"logging.warn" is deprecated and generates hundreds of warnings during testing.  warning is present in 2.7 and all of the supported 3 releases:

https://docs.python.org/2.7/library/logging.html#logging.warning


gordonmessmer:
Are there any concerns with this patch?

https://docs.python.org/3/library/logging.html#logging.warning

hroncok:
Thanks. No concerns, I just wanted the CI to re-kick after the pytest dependency change.

PR #190, opened 2019-07-09
Leave locale set to 'C' if restoring the previous one fails.

gordonmessmer:
There is already a bug filed for this issue:
https://bugs.python.org/issue30755

One perfectly reasonable response to this condition is simply to leave the locale set to 'C'.  The other would be to modify Dockerfile to install en_US support, so that we don't have to work around bugs in Python.  If you'd prefer that option, I can send a different PR.

### DIFF
--- a/pyp2rpm/utils.py
+++ b/pyp2rpm/utils.py
@@ -139,7 +139,14 @@ def c_time_locale():
     old_time_locale = locale.getlocale(locale.LC_TIME)
     locale.setlocale(locale.LC_TIME, 'C')
     yield
-    locale.setlocale(locale.LC_TIME, old_time_locale)
+    try:
+        locale.setlocale(locale.LC_TIME, old_time_locale)
+    except locale.Error:
+        # https://bugs.python.org/issue30755
+        # Python may alias the configured locale to another name, and
+        # that locale may not be installed.  In this case, the locale
+        # should simply be left in the 'C' locale.
+        pass
 
 
 def rpm_eval(macro):


pyp2rpm/utils.py
```diff
@@ -139,7 +139,14 @@ def c_time_locale():
...
     yield
-    locale.setlocale(locale.LC_TIME, old_time_locale)
+    try:
+        locale.setlocale(locale.LC_TIME, old_time_locale)
+    except locale.Error:
+        # https://bugs.python.org/issue30755
+        # Python may alias the configured locale to another name, and
+        # that locale may not be installed.  In this case, the locale
+        # should simply be left in the 'C' locale.
+        pass
```
hroncok, line 149: what if we try to at least set it to C.UTF-8 here?

gordonmessmer: That doesn't work because python thinks that "C.UTF-8" is an alias for "en_US.UTF-8" and tries to use the latter.  (Evidence: if we delete the alias then that locale is usable.)

```
>>> locale.setlocale(locale.LC_TIME, ('C', 'UTF-8'))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python3.7/locale.py", line 604, in setlocale
    return _setlocale(category, locale)
locale.Error: unsupported locale setting
>>> del(locale.locale_alias['c.utf8'])
>>> locale.setlocale(locale.LC_TIME, ('C', 'UTF-8'))
'C.UTF-8'
```

gordonmessmer: For the LC_TIME category, I don't think there would be any practical difference between "C" and "C.UTF-8" anyway, which is why I think that "pass" is a reasonable thing to do here.

hroncok: That makes the upstream bug even grosser, sigh.

gordonmessmer: I agree.  I think this is a really good illustration of why the alias is a problem, and I'll use that to improve the bug report.

gordonmessmer: Optionally, we could adopt dnf's behavior:
https://github.com/rpm-software-management/dnf/blob/master/dnf/i18n.py#L82

I know this application isn't a high priority, so I want you to know that I really appreciate the attention you've given to these patches.  Your feedback has been a tremendous help.  Once these are all merged, I plan to try to identify Fedora specs that use the ?.? glob and send them patches.

hroncok: Or remove that alias?

gordonmessmer: I tried each approach in a different branch because I think that might be the easiest way to compare them and discuss which is preferable.

Again, if this were any other category I'd probably view the value of actually fixing this differently.  However, since this is the LC_TIME category, I think there's no difference between "C" and "C.UTF-8".  locale(7) indicates that the category affects both formatting time strings and parsing time strings, but in either case I believe that only the us-english day and month names will be valid, and none of those have non-ascii characters.  The other categories remain in utf-8 encoding when only LC_TIME is set.  Those limitations are what makes me think that "pass" is fine in this specific case.

http://man7.org/linux/man-pages/man7/locale.7.html

hroncok: you have a point. I'll sleep on it

gordonmessmer: Any thoughts on this one?  I've run the tests on a private branch with this PR and #191.  The full set runs without errors (just a small number of warnings) in that branch.

hroncok: Thanks for the reminder.

PR #191, opened 2019-07-09
Remove DNF name convert tests for packages that no longer exist in rawhide

gordonmessmer:
As far as I can tell, the Fedora policy should be updated to lower-case package names.  I've updated name_convertor.py to blacklist distros that shouldn't lower-case names rather than enabling it only for mageia.  That should simplify things supposing that epel8 support will be added later.

I've replaced one fedora test set with epel7, in order to maintain tests for a distro that doesn't lower-case names.

python2-sphinx is no longer present in rawhide, so that test was removed.  python2-oslo-sphinx isn't in rawhide either, but that test can match the NameConvertor return value.  (Or we could remove it as well)


hroncok:
> As far as I can tell, the Fedora policy should be updated to lower-case package names

Based on what?

> (Or we could remove it as well)

Please remove the python2-oslo-sphinx test instead, thanks.

gordonmessmer:
> Based on what?

Oh no... I started from the fact that Fedora's python-sphinx package is lower-cased and then went to the packaging guidelines where I read "The canonical name is obtained by switching the standardized name to lower case" but that's a description of the automatic-provides behavior, not package names.  My mistake.  I'll back all that out.

gordonmessmer:
I've rebased this branch on master.  I expect the travis build to run without errors.  We'll see what happens...

hroncok:
All checks have passed.

happy dancing gif

PR #197, opened 2019-07-28
Fix repeated calls to rpm_versioned_name.

gordonmessmer:
rpm_versioned_name is called on package names multiple times in different contexts.  In the corner case of building a spec for EPEL where the base version is python3, this will lead to bad expansions:

```
Requires:       python%{python3_pkgversion}-python%{python%{python3_pkgversion}_pkgversion}-pulpcore-plugin >= 0.1rc2
Conflicts:      python%{python3_pkgversion}-python%{python%{python3_pkgversion}_pkgversion}-pulpcore-plugin >= 0.2

```

### DIFF
--- a/pyp2rpm/filters.py
+++ b/pyp2rpm/filters.py
@@ -4,7 +4,7 @@ from pyp2rpm import name_convertor
 
 def name_for_python_version(name, version, default_number=False):
     return name_convertor.NameConvertor.rpm_versioned_name(
-        name, version, default_number)
+        name, version, default_number, True)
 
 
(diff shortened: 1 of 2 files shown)


gordonmessmer:
Alternatively, rpm_versioned_name could just check for "python%{{python" in versioned_name and not change it if it is present.

gordonmessmer:
Rebased on the updated master.

gordonmessmer:
The issue that this PR fixes can be reproduced using:

```
pyp2rpm -t epel7 -b 3 -p 3 pulp_file
```

hroncok (approved):

PR #199, opened 2019-07-28
Always use the "distro" attribute for behavior checks.

gordonmessmer:
This allows users to specify a template file without requiring that its name be "distro" + ".spec".


hroncok:
Dos it preserve the old behavior?

gordonmessmer:
Yes, the output should be the same when the built-in templates are used.  When a custom template is used, the user will need to specify a distribution, whereas the current situation is that custom templates result in a failure and can't be used.

hroncok (approved):

PR #202, opened 2019-07-29
Use pypi_source macro

gordonmessmer:
PR #198 might be ... over-engineered.  This change will also support the pypi_source macro, but with fewer external requirements.  As a result, it's also a much smaller patch.

### DIFF
--- a/pyp2rpm/filters.py
+++ b/pyp2rpm/filters.py
@@ -62,10 +62,19 @@ def package_to_path(package, module):
         return package
 
 
+def macroed_url(url):
+    if (url.startswith('https://files.pythonhosted.org/packages/source/') and
+        (url.endswith('/%{pypi_name}/%{pypi_name}-%{version}.tar.gz') or
+         url.endswith('/%{pypi_name}/%{pypi_name}-%{version}.zip'))):
+        return '%{pypi_source}'
+    return url
+
+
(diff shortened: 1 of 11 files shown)


gordonmessmer:
The error here is a transient failure to download from pypi, and not a bug or unexpected change.

hroncok:
> The error here is a transient failure to download from pypi, and not a bug or unexpected change.

You should have rights to restart the build if you log in to Travis Ci with your GitHub account.

pyp2rpm/filters.py
```diff
@@ -62,10 +62,19 @@ def package_to_path(package, module):
         return package
 
 
+def macroed_url(url):
+    if (url.startswith('https://files.pythonhosted.org/packages/source/') and
+        (url.endswith('/%{pypi_name}/%{pypi_name}-%{version}.tar.gz') or
+         url.endswith('/%{pypi_name}/%{pypi_name}-%{version}.zip'))):
```
hroncok: If it ends with zip, you should return `%{pypi_source %{pypi_name} %{version} zip}` instead.

hroncok (changes requested):
There is a slight problem with the logic, but indeed this approach is better than #198

gordonmessmer:
Good catch.  Thanks!

PR #203, opened 2019-07-31
Print logging handlers to stderr on setup.py failure

gordonmessmer:


gordonmessmer:
For packages that fail to run setup.py, py2rpm will print something like:

```
Failed to extract data from setup.py script.
Check the log for details: [<FileHandler /tmp/pyp2rpm-root.log (DEBUG)>]
```

pyp2rpm/metadata_extractors.py
```diff
@@ -353,7 +353,9 @@ def _get_metadata(self, temp_dir):
                     logger.error("The JSON was: {0}".format(e.doc))
                 self.unsupported_version = current_version
         else:
-            sys.stdout.write("Failed to extract data from setup.py script.\n")
+            sys.stderr.write("Failed to extract data from setup.py script.\n")
+            sys.stderr.write("Check the log for details: {0}\n".format(
+                str(logger.parent.handlers)))
```
hroncok, line 358: Does the handler has a path attribute we could print instead?

gordonmessmer: Kind of.  We could use:
```
            sys.stderr.write("Check the log for details: {0}\n".format(
                ', '.join(x.stream.name for x in logger.parent.handlers)))
```

... but none of this is part of the documented API, as far as I know.

gordonmessmer: Result when using the --srpm flag, so that logging is sent to a file and to the console:
```
Check the log for details: /tmp/pyp2rpm-root.log, <stdout>
```

gordonmessmer: If we don't like mucking around with undocumented internals (and I don't), we could also keep a copy of output targets in a list in the logger module, and print the elements of that list in this error message.

PR #204, opened 2019-08-01
Print logging handlers to stderr on setup.py failure

gordonmessmer:


gordonmessmer:
Alternate implementation does not access "logging" module internals.

hroncok:
I like this one better.

gordonmessmer:
Is that approval?

pyp2rpm/metadata_extractors.py
```diff
@@ -353,7 +353,10 @@ def _get_metadata(self, temp_dir):
                     logger.error("The JSON was: {0}".format(e.doc))
                 self.unsupported_version = current_version
         else:
-            sys.stdout.write("Failed to extract data from setup.py script.\n")
+            import pyp2rpm.logger
```
hroncok: I don't particularly like this import here.

hroncok (changes requested):

PR #205, opened 2019-08-04
Add --skip-check option to omit test dependencies and %check section.

gordonmessmer:
Adds the --skip-check option proposed in issue #162

### DIFF
--- a/pyp2rpm/bin.py
+++ b/pyp2rpm/bin.py
@@ -105,6 +105,10 @@ class SclizeOption(click.Option):
               help='When used pyp2rpm will produce srpm instead of printing '
               'specfile into stdout.',
               is_flag=True)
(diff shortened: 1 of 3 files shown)


hroncok:
I am not very fond of this change. It promotes bad habits. If we cannot run `%check`, it shall be bconded out, not removed.

gordonmessmer:
So that I'm clear, you're suggesting that --skip-check should add a bcond, right?  Or do you mean that we shouldn't add the --skip-check option at all?

Assuming the former, should the bcond default to not running the checks?  That is, do we assume that the checks are being skipped because they won't work in mock by default (because they need network access, for example?)

hroncok:
Either not adding this at all, or it should add a bcond, disabled by default.

PR #209, opened 2019-08-18
Use latest stable version by default

gordonmessmer:
Prerelease versions are filtered out using  pip._vendor.packaging.version

### DIFF
--- a/pyp2rpm/metadata_extractors.py
+++ b/pyp2rpm/metadata_extractors.py
@@ -134,7 +134,8 @@ def venv_metadata_extension(extraction_fce):
 
         temp_dir = tempfile.mkdtemp()
         try:
-            extractor = virtualenv.VirtualEnv(self.local_file, temp_dir,
+            extractor = virtualenv.VirtualEnv(self.local_file, self.version,
+                                              temp_dir,
                                               self.name_convertor,
                                               self.base_python_version)
             data.set_from(extractor.get_venv_data, update=True)
--- a/pyp2rpm/package_getters.py
+++ b/pyp2rpm/package_getters.py
@@ -13,6 +13,7 @@ try:
     import xmlrpclib
 except ImportError:
     import xmlrpc.client as xmlrpclib
+from pip._vendor.packaging.version import parse as parse_version
 
 
 from pyp2rpm import settings
@@ -126,21 +127,27 @@ class PypiDownloader(PackageGetter):
 
     """Class for downloading the package from PyPI."""
 
-    def __init__(self, client, name, version=None, save_dir=None):
+    def __init__(self, client, name, version=None, prerelease=False,
+                 save_dir=None):
         self.client = client
         self.name = name
         try:
-            self.versions = self.client.package_releases(self.name)
+            self.versions = self.client.package_releases(self.name, True)
         except xmlrpclib.ProtocolError as e:
             sys.stderr.write("Failed to connect to server: {0} \n".format(e))
             raise SystemExit(3)
+
+        # Use only stable versions, unless --pre was specified
+        if not prerelease:
+            self.versions = [candidate for candidate in self.versions
+                             if not parse_version(candidate).is_prerelease]
 
         # If versions is empty list then there is no such package on PyPI
         if not self.versions:
             raise exceptions.NoSuchPackageException(
                 'Package "{0}" could not be found on PyPI.'.format(name))
 
-        self.version = version or self.versions[-1]
+        self.version = version or self.versions[0]
 
         # if version is specified, will check if such version exists
         if version and self.client.release_urls(name, version) == []:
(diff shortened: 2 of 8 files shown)


tests/test_virtualenv.py
```diff
@@ -69,7 +69,7 @@ class TestVirtualEnv(object):
 
     def setup_method(self, method):
         self.temp_dir = tempfile.mkdtemp()
-        self.venv = VirtualEnv(None, self.temp_dir,
+        self.venv = VirtualEnv(None, None, self.temp_dir,
```
hroncok: this starts to be hard to read, can we use kayword arguments here?

gordonmessmer:

pyp2rpm/metadata_extractors.py
```diff
@@ -134,7 +134,8 @@ def inner(self):
 
         temp_dir = tempfile.mkdtemp()
         try:
-            extractor = virtualenv.VirtualEnv(self.name, temp_dir,
+            extractor = virtualenv.VirtualEnv(self.local_file, self.version,
+                                              temp_dir,
```
decathorpe, line 138: Is there a change from the "reuse local files" PR in here?
Otherwise I don't understand the `self.name` → `self.local_file` change.
And I'm not familiar with virtualenv, but if those are positional arguments, this will blow up :)

EDIT: Ah, I see, Virtualenv is defined in this package, and you're changing it as well. Never mind.

pyp2rpm/package_getters.py
```diff
@@ -126,21 +127,27 @@ class PypiDownloader(PackageGetter):
...
+            self.versions = [candidate for candidate in self.versions
+                             if not parse_version(candidate).is_prerelease]
+
         # If versions is empty list then there is no such package on PyPI
         if not self.versions:
             raise exceptions.NoSuchPackageException(
                 'Package "{0}" could not be found on PyPI.'.format(name))
 
-        self.version = version or self.versions[-1]
+        self.version = version or self.versions[0]
```
decathorpe, line 150: Why are you using the first instead of the last version here?
Is the added `True` flag above sorting from latest to oldest?

pyp2rpm/virtualenv.py
```diff
@@ -42,7 +42,7 @@ def fill(self, path):
         '''
         self.bindir = set(os.listdir(path + 'bin/'))
         self.lib_sitepackages = set(os.listdir(glob.glob(
-            path + 'lib/python?.?/site-packages/')[0]))
+            path + 'lib/python*.*/site-packages/')[0]))
```
decathorpe, line 45: We should probably use a proper regular expression here ... `*.*` is bound to blow up somehow.

pyp2rpm/metadata_extractors.py
gordonmessmer: Right.  I'd introduced the "version" arg in this branch, so this was a result of resolving the merge conflict with the parent branch during a rebase.

pyp2rpm/package_getters.py
gordonmessmer: https://warehouse.pypa.io/api-reference/xml-rpc/
The True flag gets all releases, while only the latest release is returned otherwise.  The latest release might be a prerelease, so we need the full list if we want find the latest stable release.  I don't know how pypi behaved when "-1" was used, but I assume that the API used to return the full list by default, sorted in the opposite order.

I can add a comment to that effect.

```
>>> import xmlrpc.client as xmlrpclib
>>> client = xmlrpclib.ServerProxy('https://pypi.org/pypi')
>>> client.package_releases('pyp2rpm', True)
['3.3.3', '3.3.2', '3.3.1', '3.3.0', '3.2.3', '3.2.2', '3.2.1', '3.2.0', '3.1.3', '3.1.2', '3.1.1', '3.1.0', '3.0.2', '3.0.1', '3.0.0', '2.0.0', '1.1.2', '1.1.1', '1.1.0b', '1.1.0a', '1.0.1', '1.0.0', '0.5.2', '0.5.1', '0.5.0', '0.4.2', '0.4.1', '0.4.0', '0.3.1', '0.3.0', '0.2.0', '0.1.0']
>>> client.package_releases('pyp2rpm')
['3.3.3']
```

pyp2rpm/virtualenv.py
gordonmessmer: That string is an arg to the glob.glob() call.  We can't use regex there.  :)

This isn't strictly related to this branch.  There used to be more references to ?.?, which will break if Python 3.10 is the version after 3.9.  I missed this one in an earlier cleanup.

pyp2rpm/package_getters.py
decathorpe: Great, then that's fine.

pyp2rpm/virtualenv.py
decathorpe: I know that this is a glob.glob() call ... but it's still ugly :)
If it's solved like this everywhere else as well, then I'm fine with it.

PR #211, opened 2019-09-21
Add support for compatible releases

stephenfin:
These are a pretty nifty feature described in PEP 440 [1]. However,
they're not supported in specfiles so we need to convert them. The
approach here is rather hacky and may well break with more convoluted
requirements, but it's a good start and better than what we have right
now (nothing).

[1] https://www.python.org/dev/peps/pep-0440/#compatible-release

### DIFF
--- a/pyp2rpm/dependency_parser.py
+++ b/pyp2rpm/dependency_parser.py
@@ -58,6 +58,25 @@ def deps_from_pyp_format(requires, runtime=True):
 
     for req in requires:
         try:
+            parsed_req = Requirement.parse(req)
+            specs = parsed_req.specs[:]
+            for op, ver in parsed_req.specs:
+                # PEP-440 states that 'foo~=2.5' is equivalent to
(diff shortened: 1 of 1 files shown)


stephenfin:
Should have looked before I jumped. Seems #196 should do this and is far more mature.

gordonmessmer:
Thanks, @stephenfin .  Miro approved several of my pull requests earlier this week, and once I merge those I think we can review #196 .  I'm eager to get those done, but I spent the week preparing for vacation.  I'll probably find some down time to merge those shortly, and I hope we'll get you support for compatible-release ASAP.  :)

hroncok:
If any one of you would like to look into `=~`, the support is missing both in https://src.fedoraproject.org/rpms/pyproject-rpm-macros/blob/master/f/pyproject_buildrequires.py and https://github.com/rpm-software-management/rpm/blob/master/scripts/pythondistdeps.py#L244

gordonmessmer:
@stephenfin I've updated #196 if you'd like to test it and make any suggestions.

PR #212, opened 2019-09-26
Specify distro for new Convertor test.

gordonmessmer:
PR 195 was merged, but did not have the "distro" argument in its tests, which is required since the merge of PR 199.  This change fixes that deficiency.

### DIFF
--- a/tests/test_convertor.py
+++ b/tests/test_convertor.py
@@ -102,7 +102,8 @@ class TestConvertor(object):
     def test_merge_versions_epel7(self, self_bv, self_pv, data_pv,
                                   expected_bv, expected_pv):
         c = Convertor(package='pkg', base_python_version=self_bv,
-                      python_versions=self_pv, template='epel7.spec')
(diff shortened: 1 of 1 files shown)


hroncok:
I don't mind the change, but the tests blew up about spec2scl.

gordonmessmer:
I noticed that.  I don't think it's related to changes in this project.  It's probably the result of some change in spec2scl dropping support for py27.  I'll look into it further as soon as I can.  Since we plan to stop testing py27 anyway, we can probably just specify a maximum version for that package temporarily, until we get around to removing those tests.

gordonmessmer:
The problem occurred in a couple of different Travis builds, but I was unable to reproduce it in my own environment today.  I restarted the Travis build, and it doesn't look like it's happening now.

gordonmessmer:
This change should resolve the test failures in 206 and 209.

hroncok (approved):

PR #213, opened 2019-11-11
allow skipping extras_requires from (Build)Requires

evgeni:
sometimes one wants to generate a "minimal" RPM, with no extra
dependencies, which this should help with.


gordonmessmer:
Glancing over this, the only thing that stands out is that there are no tests.  Could you identify a module that has extra requires and include a test case with and without them?

evgeni:
Yes, I will add tests this week.

evgeni:
@gordonmessmer how is this? I've only added tests for runtime deps as that's what I had at hand in a package of mine. Should I search for a package that does the same in build deps?

evgeni:
nvm, found one

PR #215, opened 2020-01-29
added blackPanther OS support

blackPantherOS:
We added blackPanther support to your the beta, please merge if you agree on this patch


Conan-Kudo:
@blackPantherOS Is there something wrong with the Mageia template for your use? The only difference is whether Python 2 is enabled by default or not...

ghost:
> @blackPantherOS Is there something wrong with the Mageia template for your use? The only difference is whether Python 2 is enabled by default or not...

The python2 test used wrong pyparsing. We had to add the pyparsing to the deps in the tox.ini file.

gordonmessmer:
I would ask the same as @Conan-Kudo : What does this PR allow you to do that you could not do with `pyp2rpm -o mageia -b2 -p3` ?

pyp2rpm/settings.py
```diff
@@ -1,17 +1,18 @@
 from pyp2rpm import utils
 
-DEFAULT_TEMPLATE = 'fedora'
+DEFAULT_TEMPLATE = 'blackpantheros'
```
gordonmessmer, line 3: At this point, it looks like you are using this branch as a general-purpose development branch, no longer limited to changes that fit the PR.

Please open a new PR if you have changes that would be beneficial to users of this project, especially features that do not already exist with the appropriate command-line arguments.

Issue #223, opened 2020-03-08
Testing is broken by external changes


gordonmessmer:
Several changes in external tools have broken the test suite.

Issue #182, opened 2019-03-29
Compatible release operator ~= in dependencies yield an RPM error


mikedep333:
After running `pyp2rpm -s pulp_file`, I got this line:

> BuildRequires:  python3dist(pulpcore-plugin) ~= 0.1rc1
`rpmbuild -bs` failed with:
error: line 15: Dependency tokens must begin with alpha-numeric, '_' or '/': BuildRequires:  python3dist(pulpcore-plugin) ~= 0.1rc1

I worked around this manually by changing `~=` to `>=`

hroncok:
Another elif in here might solve this.

https://github.com/fedora-python/pyp2rpm/blob/438a987c4221f4590961d6bf449dc0b6d3c31266/pyp2rpm/dependency_parser.py#L28-L37

What is the exact meaning of `~=`?

hroncok:
Also not thet in this case, a proper  `0.1` release won't be `>= 0.1rc1`, given the way how rpm sorts:

```
$ rpmdev-vercmp 0.1 0.1rc1
0.1 < 0.1rc1
```

gordonmessmer:
This should be resolved in v.3.3.4.  Please re-open if you can reproduce the issue.

PR #224, opened 2020-03-08
Fix tests

gordonmessmer:
Some of these changes are temporary.  I'd like to merge them in and close a few issues for one more release before removing support for EOL python versions, at which point we can push a new major version.

### DIFF
--- a/tests/test_metadata_extractors.py
+++ b/tests/test_metadata_extractors.py
@@ -311,7 +311,7 @@ class TestSetupPyMetadataExtractor(object):
                            ['BuildRequires', 'python-setuptools'],
                            ['BuildRequires', 'python-sphinx']]),
         (1, 'py_modules', ['pytest']),
-        (1, 'packages', ['_pytest']),
+        # (1, 'packages', ['_pytest', 'pytest']), # py34 will get setuptools 43, and packages will not have 'pytest'
         (1, 'home_page', 'http://pytest.org'),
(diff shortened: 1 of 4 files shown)


gordonmessmer:
@decathorpe Can you look at the PR I just opened for issue #223 ?

tests/test_metadata_extractors.py
```diff
@@ -311,7 +311,7 @@ def setup_method(self, method):
                            ['BuildRequires', 'python-setuptools'],
                            ['BuildRequires', 'python-sphinx']]),
         (1, 'py_modules', ['pytest']),
-        (1, 'packages', ['_pytest']),
+        # (1, 'packages', ['_pytest', 'pytest']), # py34 will get setuptools 43, and packages will not have 'pytest'
```
decathorpe: Can you conditionally enable/disable this test depending on the python version?

decathorpe:
Other than the inline comment, looks fine to me ... though I'm wondering if it's actually still necessary to test python2 code.

tests/test_metadata_extractors.py
gordonmessmer: I've pushed a change to do that.

gordonmessmer:
I'd like to continue testing py27 until issue #182 is resolved, for the sake of platforms that still include and support it.  Once we publish at least one version with support for =~, we can remove support for EOL python versions and rev the major.

decathorpe:
Looks good to me now. Thanks

PR #226, opened 2020-03-10
Use local file for virtualenv rather than downloading another copy

gordonmessmer:


gordonmessmer:
As an added benefit, the test suite runs a good deal faster. When pypi is slow, the test suite would sometimes exceed Travis CI's maximum run time.

gordonmessmer:

This is a trivial, one-line change, and reverting one of the test changes.  In the course of fixing #209, I figured out that the virtualenv module was downloading an extra copy of the module, which is why we were getting inconsistent results from pytest.  For some platforms, we were getting information about a prerelease of pytest on pypi.

hroncok (approved):

PR #229, opened 2020-03-23
Bump version to 3.3.4 and update changelog

gordonmessmer:


gordonmessmer:
@decathorpe This seems like a good time to publish a new release, and then remove support for WHL archives and Python < 3.6.  Are there any other issues you think should be addressed before that?

decathorpe:
I don't think so ... but you definitely have a better overview of open problems than I do :)

hroncok:
Do it

PR #232, opened 2020-04-19
Discontinue removing bundled egg-info.

gordonmessmer:


hroncok:
Is `has_bundled_egg_info` unused now?

The proper fixer might be to only do this if there are binaries inside. However that might be tedious.

I strongly suggest to put the rationale into the commit message. The lack of rationale for adding this in the first place is now hurting. 8 years from now, somebody might wonder why was this removed and we could be long gone.

gordonmessmer:
> Is has_bundled_egg_info unused now?

Seems to be.  I waffled on removing it, but I suppose there wouldn't be a reason not to... We can always pull it back from history.

> The proper fixer might be to only do this if there are binaries inside. However that might be tedious.

I can look into that.  Do you happen to know an example of a package that's published with binaries inside?

> I strongly suggest to put the rationale into the commit message. 

Definitely.

hroncok:
> Do you happen to know an example of a package that's published with binaries inside?

I don't, really. Let's say we don't do this until proven otherwise?

gordonmessmer:
Reference, in case anyone reviews this in the future:

https://bugzilla.redhat.com/show_bug.cgi?id=1825455#c2

PR #233, opened 2020-04-19
Handle all-zero versions without crashing.

gordonmessmer:


hroncok:
That makes sense. Is this for https://github.com/rpm-software-management/rpm/issues/1183 ?

torsava:
Looks good. I think a new test would be in order too.

torsava:
I'm adding a test to pythondistdeps script for this. So if we have a package version `0+unknown`, is the correct thing to do to ignore the local version and just evaluate to RPM version `0`? That's what the pythondistdeps script does currently.

@gordonmessmer @hroncok

hroncok:
I think so, yes. The local version thing is described in https://www.python.org/dev/peps/pep-0440/#local-version-identifiers Most importantly:

```
Local version labels have no specific semantics assigned, but some syntactic restrictions are imposed.
```

However it also says:

> Comparison and ordering of local versions considers each segment of the local version (divided by a .) separately. If a segment consists entirely of ASCII digits then that section should be considered an integer for comparison purposes and if a segment contains any ASCII letters then that segment is compared lexicographically with case insensitivity. When comparing a numeric and lexicographic segment, the numeric section always compares as greater than the lexicographic segment. Additionally a local version with a great number of segments will always compare as greater than a local version with fewer segments, as long as the shorter local version's segments match the beginning of the longer local version's segments exactly.

Which seems very complicated to check against how rpm behaves.

I am fine if we deliberately document and say that it is simply omitted. Consider also:

> Local version identifiers SHOULD NOT be used when publishing upstream projects to a public index server...

If desired, we might convert it to `^` suffix, but meh.

gordonmessmer:
@decathorpe

tests/test_dependency_parser.py
```diff
@@ -54,6 +54,9 @@ class TestDependencyParser():
         ('pyp2rpm~=0.9.3.1', True, True,
          [['Requires', 'pyp2rpm', '({name} >= 0.9.3.1 with {name} < 0.9.4)']]
          ),
+        ('nb2plots>0+unknown', True, True,
+         [['Requires', 'nb2plots', '{name} > 0']]
```
decathorpe, line 58: Shouldn't this be `>= 0`, if the `+unknown` part is ignored in generators?
Or maybe just drop the Version restriction and require only plain `nb2plots`, since all versions (except 0~foo) satisfy `>= 0`?

gordonmessmer: Since "+unknown", the original requirement is simplified to "nb2plots>0", and the function output is "nb2plots > 0".  

IIRC, this requirement was used in a package, which is how I noticed that the conversion routines crash if the version required is all zeros.

decathorpe: Okay, if this works for real world examples, then LGTM.

PR #240, opened 2020-07-03
Sync python dependency conversion with pyreq2rpm.

gordonmessmer:


gordonmessmer:
@decathorpe

decathorpe:
Does this PR conflict with the "do not crash on all-zero version" PR? Otherwise, :shipit:

PR #242, opened 2020-10-29
adhere to the new pypi.org rate-limiting

evgeni:
pypi.org has rate-limited their XML-RPC API to 1 request/second, see
https://github.com/pypa/warehouse/issues/8753 for details.

Let's adhere to that limit until we migrate to the new JSON API.

### DIFF
--- a/pyp2rpm/convertor.py
+++ b/pyp2rpm/convertor.py
@@ -1,7 +1,10 @@
+__metaclass__ = type
+
 import logging
 import os
 import re
 import sys
+import time
 try:
     import urllib2 as urllib
 except ImportError:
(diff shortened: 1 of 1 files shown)


evgeni:
The test failures on Python2 are related, but my motivation to fix Python2 issues is ∅

gordonmessmer:
It looks like the tests are failing on all python versions, not just python 2.

I'm also seeing the HTTPTooManyRequests fault in the Python 3 versions, which is odd because testing your change in isolation suggests that it should be rate limiting correctly.

I got a bug report via redhat bugzilla just before the PR was opened, and I'm working on a similar fix.  I'll see if I can figure out why this isn't working as expected and take the more readable approach.

gordonmessmer:
I've pushed some code to the issue_246 branch.  It might be faster to switch to the JSON API than to make the XMLRPC interface work again.

gordonmessmer:
This issue is resolved in v3.3.5

evgeni:
Thanks!

PR #244, opened 2020-10-30
 Handle PyPI rate limit of one request per second. 

gordonmessmer:
https://bugzilla.redhat.com/show_bug.cgi?id=1892478

### DIFF
--- a/pyp2rpm/convertor.py
+++ b/pyp2rpm/convertor.py
@@ -314,11 +314,15 @@ class RateLimitedServerProxy(xmlrpclib.ServerProxy):
     """Handle rate limit at PyPI."""
 
     def __getattr__(self, name):
-        def slow_call(*args):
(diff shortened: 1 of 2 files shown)


jamescassell:
Thanks for the fix. I assume this is why COPR is failing python PyPI builds. RH bug is private, unfortunately, otherwise I'd subscribe.

gordonmessmer:
@jamescassell  Do you have a link to the logs of any of those failed builds?

jamescassell:
This PR works around the issue for me when running it by hand. The COPR error was a different message IIRC.

gordonmessmer:
This issue is resolved in v3.3.5

gordonmessmer:
... but do let me know if the COPR errors look like they're related to a pyp2rpm problem.

PR #245, opened 2020-10-30
Don't process PyPI version list if user specifies a version.

gordonmessmer:


pyp2rpm/package_getters.py
```diff
@@ -127,25 +127,28 @@ def __init__(self, client, name, version=None, prerelease=False,
...
         self.name = name
-        self.versions = self.client.package_releases(self.name, True)
-
-        # Use only stable versions, unless --pre was specified
-        if not prerelease:
-            self.versions = [candidate for candidate in self.versions
-                             if not parse_version(candidate).is_prerelease]
+        if version:
+            # if version is specified, will check if such version exists
+            if self.client.release_urls(name, version) == []:
```
evgeni: I think `if not self.client.release_urls(…):` is more readable (as empty lists are False-y).

evgeni (approved):
Reading the code it should fix my issue!

gordonmessmer:
@decathorpe

decathorpe:
Makes sense. If it works, :shipit:

PR #247, opened 2020-11-03
Replace XMLRPC client with JSON API client.

gordonmessmer:


gordonmessmer:

thebeanogamer:
I've been able to use this to create a SPEC where I couldn't with the version on `master`, so I think this is working.

decathorpe:
@gordonmessmer I have never used an XMLRPC client before and I'm not familiar with the PyPI API, so I'd rather somebody else review your changes. Sorry :/

gordonmessmer:
If you're wiling to discuss the patch, I think your familiarity with the PyPI API should not be a roadblock.  As review goes, my primary concern is that at least one other person is aware of changes being made to the application, and to ensure that if a co-maintainer would like to see more comments or more idiomatic code, that those changes are made before merging.

If not, I'll see who else might be available.  Thanks. :)

pyp2rpm/convertor.py
```diff
@@ -281,43 +283,57 @@ def metadata_extractor(self):
...
     def client(self):
-        """XMLRPC client for PyPI. Always returns the same instance.
+        """JSON client for PyPI. Always returns the same instance.
 
         If the package is provided as a path to compressed source file,
         PyPI will not be used and the client will not be instantiated.
 
         Returns:
-            XMLRPC client for PyPI or None.
+            JSON client for PyPI or None.
```
hroncok, line 292: When does this return None?

gordonmessmer: If pyp2rpm is called on a local archive file, the pypi client won't be used.

hroncok: Right. When `self.pypi` is false.

gordonmessmer: Is there an implication that I'm missing?

hroncok: No. It was just a but confusing to me, so I've asked. But it is not wrong.

pyp2rpm/convertor.py
```diff
@@ -4,17 +4,16 @@
 import sys
 try:
     import urllib2 as urllib
+    from urllib2 import urlopen, HTTPError, URLError
 except ImportError:
     import urllib
     from urllib import request
+    from urllib.request import urlopen, HTTPError, URLError
```
gordonmessmer, line 11: These imports are an excellent target for improvement when python 2 support is removed.

pyp2rpm/convertor.py
```diff
@@ -281,43 +283,57 @@ def metadata_extractor(self):
...
-            XMLRPC client for PyPI or None.
+            JSON client for PyPI or None.
         """
-        if self.proxy:
-            proxyhandler = urllib.ProxyHandler({"http": self.proxy})
-            opener = urllib.build_opener(proxyhandler)
-            urllib.install_opener(opener)
-            transport = ProxyTransport()
-        if not hasattr(self, '_client'):
-            transport = None
```
gordonmessmer, line 298: Proxy handling was moved to the PyPIClient class.  This should fix proxy support that may be currently broken because transport is set to None immediately after constructing it.

pyp2rpm/package_getters.py
```diff
@@ -35,7 +31,7 @@ def get_url(client, name, version, wheel=False, hashed_format=False):
             client, name, version))
         raise exceptions.MissingUrlException(
             "Some kind of error while communicating with client: {0}.".format(
-                client), exc_info=True)
```
gordonmessmer, line 38: exc_info is an argument for logger functions, not Exceptions.

pyp2rpm/package_getters.py
```diff
@@ -131,11 +127,7 @@ def __init__(self, client, name, version=None, prerelease=False,
                  save_dir=None):
         self.client = client
         self.name = name
-        try:
-            self.versions = self.client.package_releases(self.name, True)
-        except xmlrpclib.ProtocolError as e:
-            sys.stderr.write("Failed to connect to server: {0} \n".format(e))
-            raise SystemExit(3)
```
gordonmessmer, line 138: Connection failure exception was moved for the PyPIClient class.

pyp2rpm/convertor.py
```diff
@@ -281,43 +283,57 @@ def metadata_extractor(self):
...
+        return self.cache[(name, version)]
+
+    def release_data(self, name, version):
+        return self.get_json(name, version)['info']
+
+    def release_urls(self, name, version):
+        return self.get_json(name, version)['urls']
+
+    def package_releases(self, name, show_hidden):
+        return self.get_json(name, None)['releases'].keys()
```
gordonmessmer, line 339: The JSON response contains the same data structures that were provided by XMLRPC responses on the older API.  release_data, release_urls, and package_releases were the XML-RPC methods previously used, so the new client class simply implements those methods to avoid any restructuring of the calls outside of this class.

gordonmessmer:
@hroncok If this all looks good to you, I'll publish a new release.

hroncok:
It fixes the problem. Ship it!

PR #253, opened 2020-12-29
Improve logging of rpmbuild output.

gordonmessmer:


pyp2rpm/bin.py
```diff
@@ -246,7 +246,13 @@ def main(package, v, prerelease, d, s, r, proxy, srpm, p, b, o, t, venv, autonc,
 
         if srpm:
             msg = utils.build_srpm(spec_path, d)
-            logger.info(msg)
+            if isinstance(msg, bytes):
+                for line in msg.decode().split('\n'):
```
hroncok: This should be more robust:

```suggestion
                for line in msg.decode('utf-8', errors='replace').split('\n'):
```

gordonmessmer: Good point.

gordonmessmer:

pyp2rpm/bin.py
```diff
@@ -246,7 +247,14 @@ def main(package, v, prerelease, d, s, r, proxy, srpm, p, b, o, t, venv, autonc,
 
         if srpm:
             msg = utils.build_srpm(spec_path, d)
-            logger.info(msg)
+            if isinstance(msg, bytes):
+                for line in msg.decode(locale.getpreferredencoding(),
```
decathorpe, line 251: Why use locale.getpreferredencoding instead of just 'utf-8'? I assume that on systems that don't use a UTF-8 locale all sorts of other things will go wrong too.

gordonmessmer: locale.getpreferredencoding() is consistent with other uses of decode() in the application (as in archive.py and package_data.py), and will return the encoding indicated by LANG or LC_CTYPE.  Since we're decoding the output of the "rpmbuild" command, I would expect its output to be in the encoding specified by one of those variables.

decathorpe: Fair enough. then LGTM

PR #255, opened 2021-01-01
Remove "version" arg from VirtualEnv class init.

gordonmessmer:
Remove "version" arg from VirtualEnv class init.  Because "name"
is a local file, "version" is not necessary.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ MANIFEST
 .eggs/
 tests/test_output/
 tests/test_data/utest-*.tar.gz
+tests/test_data/*/*.egg-info
--- a/pyp2rpm/metadata_extractors.py
+++ b/pyp2rpm/metadata_extractors.py
@@ -134,7 +134,7 @@ def venv_metadata_extension(extraction_fce):
 
         temp_dir = tempfile.mkdtemp()
         try:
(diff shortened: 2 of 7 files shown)


gordonmessmer:

I introduced a bug in #209 and #226 .  Since pyp2rpm builds a virtualenv from a local file, it doesn't need to specify a version.  The version spec is incompatible with specifying a local file, which causes setup to fail, and no data can be extracted from virtualenvs.  None of the tests specifically required virtualenv data, so I've added additional tests on virtualenv specifically.

There's still one issue that the test reveals, which is that pyp2rpm will try to use python3 for some tests, even when it is running in the Python 2.7 virtualenv created by tox.  Rather than spend time on that, I think I might just file that under "known issues" temporarily until Python 2.7 support is removed.

Travis tests aren't running most of the time due to docker pull rate limits.  All tests for this branch are successful except for one test under the py27 env specifically:
tests/test_virtualenv.py::TestVirtualEnvGetData::test_get_data[/home/gordon/git/pyp2rpm/tests/test_data/utest-0.1.0.tar.gz-expected0] FAILED

decathorpe:
If "name" in VirtualEnv constructors is really always a local file, then this LGTM.

PR #260, opened 2021-01-11
Include rc/post/dev in version.

gordonmessmer:


gordonmessmer:

Overall: rather than discarding a/b/rc/dev/post information provided in Python versions, those will be preserved and transformed into something that rpm understands.  Because the version string may be modified for the rpm Version:, but used literally in filesystem paths, the original value needs to be stored in a define/global value in the spec file, the same way that package names are handled.  Most values of %{version} are replaced with %{pypi_version}, similar to the use of %{pypi_name}

In filters.py, I've introduced two new filters.  One will convert a/b/rc/dev to a tilde suffix which should be supported on RHEL 6 and 7 (I'll test that asap), and the other will convert a/b/rc/dev to a tilde suffix and also convert post to a caret suffix.  If the original version doesn't need any transformation, the filter returns "%{pypi_version}", which produces a spec file that needs minimal modification for future releases.  The exception is for the changelog, where the pypi_version macro can't be used, so the filters accept a flag indicating whether or not the macro or the literal version is returned.

Also in filters, macroed_url is updated to recognize the use of pypi_version rather than version.

package_getters.py is updated to preserve the version suffix which was previously discarded.

The majority of this PR is changes to the templates and the tests.

gordonmessmer:
CentOS 7 can install packages with ~ version, and:

```
# rpmdev-vercmp 1.9 1.9.rc1
1.9 < 1.9.rc1
# rpmdev-vercmp 1.9 1.9~rc1
1.9 > 1.9~rc1

```

decathorpe:
This is a bit above my paygrade ...

The thing that*s weird though is that this PR has six commits of which 5 have the same message, making it hard to review, since you can't tell which commit is supposed to do what. Breaking the "flesh" of this change into multiple commits would also make it easier to review :)

gordonmessmer:
I'd intended to squash it all to one since the vast majority of this change is just the same update to many templates and the corresponding test spec files.  I've reorganized the commits.

decathorpe:
Thanks! I'll leave some comments inline.

pyp2rpm/filters.py
```diff
@@ -71,11 +74,49 @@ def macroed_url(url):
...
+    else:
+        default = version
+    re_match = re.compile(
+        r"(\d+\.?\d*\.?\d*\.?\d*)\.?((?:a|b|rc|dev)\d*)").search(
+            version)
+    if not re_match:
+        return default
+    rpm_version, rpm_suffix = re_match.groups()
+    if rpm_suffix.startswith("dev"):
+        return '{}~~{}'.format(rpm_version, rpm_suffix)
```
decathorpe, line 95: Do you think we actually *should* support -dev versions? Doing so makes this non-standard double-tilde necessary, and it's at least not covered by the Fedora packaging guidelines at all.

pyp2rpm/filters.py
```diff
@@ -71,11 +74,49 @@ def macroed_url(url):
...
+        return '{}~{}'.format(rpm_version, rpm_suffix)
+
+
+def rpm_version(version, use_macro=True):
+    if use_macro:
+        default = '%{pypi_version}'
+    else:
+        default = version
+    re_match = re.compile(
+        r"(\d+\.?\d*\.?\d*\.?\d*)\.?((?:a|b|rc|post|dev)\d*)").search(
```
decathorpe, line 116: Why does this regex contain `post` as choice, but the one above does not?
And should we actually support `post` and `dev` versions in pyp2rpm?

decathorpe:
Last question: When running pyp2rpm now, will it choose the latest version regardless of alpha / beta / rc status now? At least I think by default it should use the latest "stable" release, and using a newer prerelease should be optional (either by specifying the version manually or by setting a `--use-prereleases` flag or similar).

hroncok:
https://github.com/fedora-python/pyp2rpm/blob/24002cc7df516332bf2f4198cb145d54c8ce3c98/pyp2rpm/bin.py#L127-L130

pyp2rpm/filters.py
hroncok: The double tilde is what the Python dist RPM dependency generator does for dev.

pyp2rpm/filters.py
```diff
@@ -71,11 +74,49 @@ def macroed_url(url):
     return url
 
 
+def rpm_version_410(version, use_macro=True):
+    if use_macro:
+        default = '%{pypi_version}'
```
hroncok, line 85: Where is this from?

hroncok:
Oh. So I was thinking. What if we have a `%{pypi_version}` macro in Fedora and EPEL?

Usage:

```
Version: 7.5.0~~dev1
%{pypi_version} -> 7.5.0dev1

Version: 7.5.0~b4
%{pypi_version} -> 7.5.0b4

Version: 7.5.0
%{pypi_version} -> 7.5.0

Version: 7.5.0^post1
%{pypi_version} -> 7.5.0post1


%{pypi_version 7.5.0~~dev1} -> 7.5.0dev1
%{pypi_version 7.5.0~b4} -> 7.5.0b4
%{pypi_version 7.5.0} -> 7.5.0
%{pypi_version 7.5.0^post1} -> 7.5.0post1
```

I would definitively use it in some of the packages. There is `%{version_nodots}` from rust srpm macros, but it adds `-` by default :/ See also https://github.com/rpm-software-management/rpm/issues/1219

pyp2rpm/filters.py
hroncok: Oh, it is defined in the spec file.

pyp2rpm/filters.py
gordonmessmer: I've added documentation to the filter functions.  rpm 4.10 - 4.14 support pre-release suffixes.  4.15 added support for a post-release suffix.

gordonmessmer:
> Oh. So I was thinking. What if we have a `%{pypi_version}` macro in Fedora and EPEL?

From a maintenance point of view: I lean toward consistency with existing behavior.  pyp2rpm already uses %{pypi_name} for the unmodified name, and converts to an rpm-semantics name where those are required.  If we do the inverse here, storing the converted version in the Version field and converting to the pypi semantics for the filesystem paths, we should have a good reason for the inconsistent behavior.

From an implementation point of view: Python is a bit permissive with version parsing, and I worry that someday we'd see packages where attempting to convert *back* to the original version didn't work (as it would not have in issue 259, which prompted this change).

```
>>> from pkg_resources import parse_version
>>> parse_version('0.1.post8')
<Version('0.1.post8')>
>>> parse_version('0.1post8')
<Version('0.1.post8')>

```

hroncok:
I see. I'd still like to have macro I can use that would strip tildes and carets from version :(

decathorpe:
> I see. I'd still like to have macro I can use that would strip tildes and carets from version :(

The Rust macros have some version of this, at least for stripping tilde to convert back to semver (`%{version_no_tilde}`).

hroncok:
Yes but that replaces the tilde with dash by default and the only way to wwork around it is to provide empty string as the first argument to that macro. Which is not readable: `%{version_no_tilde %{quote:}}`

gordonmessmer:
@decathorpe

gordonmessmer:
> I see. I'd still like to have macro I can use that would strip tildes and carets from version :(

https://src.fedoraproject.org/rpms/python-rpm-macros/pull-request/101

decathorpe:
I think I already reviewed this once ... unless you made some drastic changes, and it works ... :shipit:

PR #264, opened 2021-03-24
Ensure that utest and JSON test data are included in sdist.

gordonmessmer:


gordonmessmer:
@decathorpe

decathorpe:
If it works as expected: :shipit:

setup.py
```diff
@@ -13,7 +14,22 @@
...
 provide that much information though)."""
 
+class build_py(_build_py):
+    def run(self):
+        # Run the normal build process
+        _build_py.run(self)
+        # Build test data
+        from subprocess import call
+        from shutil import copy
+        call(['python3', 'setup.py', 'sdist'],
```
hroncok, line 24: I realize I am a bit late here, but should this be `sys.executable` instead of `'python3'`?

PR #267, opened 2021-05-08
Wait for rpmdev-setuptree to exit.

### LINKED ISSUE #266: pyp2rpm doesn't work when build tree is not initialized
Hello. It looks like 
$ pyp2rpm ./dist/tmt-provision-minute-0.1.0.tar.gz > tmt-provision-minute.spec

tar.gz file is generated by:
python3 -m build

command is unable to proceed when build tree (~/rpmbuild/{BUILD,RPMS,SOURCES,SPECS,SRPMS}) is not initialized.

At first run, it generates empty spec file.
After second run, everything works fine.

Version:
pyp2rpm-3.3.6-1.fc33.noarch

Workaround:
call rpmdev-setuptree before calling pyp2rpm.

gordonmessmer:
Fixes #266

### DIFF
--- a/pyp2rpm/package_getters.py
+++ b/pyp2rpm/package_getters.py
@@ -104,7 +104,7 @@ class PackageGetter(object):
             else:
                 try:
                     subprocess.Popen(
-                        'rpmdev-setuptree', stdout=subprocess.PIPE)
+                        'rpmdev-setuptree', stdout=subprocess.PIPE).wait()
                     logger.info("Using rpmdevtools package to make rpmbuild "
                                 "folders tree.")
                 except OSError:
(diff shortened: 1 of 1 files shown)


gordonmessmer:
@decathorpe I'd like to publish a version to support py3.10, and ship a handful of fixes with that.  If you have time to glance at a few of these, that'd be great.  (I'll ping you on the others.)  If there are no objections, I'll merge these next weekend.

decathorpe:
Looks good to me. Though the whole thing could be replaced with a higher level subprocess.run call instead.

gordonmessmer:
Run was added in 3.5, according to the docs.  At some point, we'll raise the minimum version, but right now it seems like there are still a lot of EPEL users, so I keep putting that off.

PR #268, opened 2021-06-14
Replace Travis CI with GitHub actions.

gordonmessmer:


gordonmessmer:
There are a few test fixes included as well.

I was never able to complete the move to travis-ci.com, probably due to access issues, but their documentation isn't very helpful in sorting that out.  But, we could just move to GitHub Actions instead and call it a day.  It might also be a good idea to review my access to the fedora-python group and check that I have no more access than necessary following changes when we were working on the travis-ci.com move.

What do you think, @hroncok ?

hroncok:
+100 for moving from Travis CI to GitHub actions. I used to be a big advocate for Travis CI, but the current state of the service is not good.

As for the access -- I am more than happy to have you in the GitHub organization.

PR #269, opened 2021-07-03
Bump version to 3.3.7 and update changelog

gordonmessmer:


gordonmessmer:
@hroncok I'm assuming there's no reason to include the .gitignore (or .github) files in the sdist.  Any thoughts on excluding those in MANIFEST.in?

hroncok:
I don't understand why they are included in the first place, but yeah, I'd exclude them if they are.

PR #273, opened 2021-10-03
Use major and minor version for python 3 sphinx-build on EPEL 7.

gordonmessmer:


nkadel:
Please double check the expired failures and merge this.

PR #277, opened 2021-12-22
Overhaul tests

### LINKED ISSUE #275: As more python2 modules are retired from rawhide, the epel dnf test results have changed
Several tests are no longer producing the expected result as a result of a change in the available packages in the repositories.  For example, python-MarkupSafe should become pythonX-markupsafe after a dnf nc lookup, but when rendering an epel template using the rawhide repos as reference and python2 as the base version, the dnf lookup produces no result.

gordonmessmer:
A number of tests needed to be adjusted as Python2 is being retired in Fedora.  

Resolves #275


tests/test_name_convertor.py
```diff
@@ -58,12 +58,12 @@ def setup_method(self, method):
         self.dnc = DandifiedNameConvertor('fedora')
 
     @pytest.mark.parametrize(('pypi_name', 'version', 'expected'), [
-        ('Babel', '2', 'python2-babel'),
-        ('Babel', '3', 'python3-babel'),
-        ('MarkupSafe', '2', 'python2-markupsafe'),
-        ('MarkupSafe', '3', 'python3-markupsafe'),
-        ('Jinja2', '2', 'python2-jinja2'),
-        ('Jinja2', '3', 'python3-jinja2'),
+        ('Babel', '2', 'babel'), # Bad result; Present in repo
```
hroncok, line 61: I don't understand this value.

Dockerfile
```diff
@@ -8,8 +8,8 @@ LABEL summary="Image for running automatic tests of pyp2rpm in Travis CI" \
       name="pyp2rpm-tests" \
       maintainer="Michal Cyprian <mcyprian@redhat.com>"
 
-RUN INSTALL_PKGS="gcc tox python27 python34 python35 python36 python37\
-                  python2-setuptools python3-setuptools" && \
+RUN INSTALL_PKGS="gcc tox python36 python37 python38 python39 python310 \
```
hroncok:
```suggestion
RUN INSTALL_PKGS="gcc tox python3.6 python3.7 python3.8 python3.9 python3.10 \
```

tests/test_name_convertor.py
gordonmessmer: If the dnf python module is available, and if the AutoProvidesNameConvertor isn't used, then DandifiedNameConvertor will search the repos for a package name that best fits the criteria of python version and package name.  In this case, the package name in Fedora's repos that best matches the module name "Babel" is "babel, now that "python2-babel" has been retired.

During rendering of the template, that name is transformed again using the base NameConvertor class.

All of this seems very complex to me, and I find it confusing every time I have to interact with it, but simplifying the design is out of scope for fixing the tests, I think.  ~~The system does work, it's just difficult to understand.~~

(I want to walk that back a bit, because it sounds like I'm defending DandifiedNameConvertor, when in fact I wonder whether it makes any sense to verify that "python2-babel" is not in the repos, only to later generate that as the name of a dependency.  Nor does it really make sense to use dnf to generate names for any distribution other than the one whose repos are queried.  Name conversion as a whole probably should be reworked, but again, I think that's out of scope for this change.)

hroncok: The thing I don't understand is that the babel package in the repos isn't the Python 2 version of babel.

gordonmessmer: Right.  The DandifiedNameConvertor is just looking for whatever package most closely matches the name of the requirement.  It doesn't determine whether the result actually provides a python module, at all.  :shrug:

hroncok: That is horrible :scream: 

Either way, if that is the intended behavior, it shall be tested, but I suggest we make a note not to do that.

hroncok (approved):

tests/test_name_convertor.py
gordonmessmer: > I suggest we make a note not to do that.

Let's call that issue #278

PR #282, opened 2022-03-23
DOC: README: markdown, syntax, quoting

westurner:
- [x] DOC: wrap the docker commands in a markdown fenced code block
- [x] DOC,SEC: double-quoting `"$(pwd)"`, which can contain spaces (and should probably be called once and cached, though the output of `pwd` is unlikely to change between invocations)

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,25 +4,29 @@
 pyp2rpm
 =======
 
-A tool to convert a PyPI package to an RPM SPECFILE or to generate an SRPM.
+A tool to convert a PyPI package to an RPM `SPECFILE` or to generate an SRPM.
 Under heavy development, see the TODO file for a list of planned features.
 pyp2rpm currently ships with Fedora and Mageia specific templates.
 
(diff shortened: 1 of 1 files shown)


westurner:
- [ ] DOC: README: Reference SRPM and `SPECFILE` for pyp2rpm itself:
  - https://src.fedoraproject.org/rpms/pyp2rpm/blob/rawhide/f/sources # SHA-512, no URL
  -   https://src.fedoraproject.org/rpms/pyp2rpm/blob/rawhide/f/pyp2rpm.spec
- [ ] DOC: README: Implementers heading?
  - Fedora COPR auto-generates _ w/ pyp2rpm: https://docs.pagure.org/copr.copr/user_documentation.html
  - https://libraries.io/pypi/pyp2rpm
    - "Used by"
      - https://libraries.io/github/mcyprian/pyp2rpm_analyzer

westurner:
Thanks!

PR #289, opened 2022-12-16
Add RHEL 8 and RHEL 9, discard RHEL 6 special options

nkadel:
Stop publishing python2 builds for RHEL 6 and 7. Add python3 builds for RHEL 8 and RHEL 9.

RHEL 6 is deprecated, and before its discard had support for python3 form EPEL before its obsolescence.


nkadel:
Is anyone maintaining this at all? I hesitate to do it myself, because I debug much better than I add new features in python. But if needed to get commits like this merged, I could take it.

PR #301, opened 2026-04-21
security: Escape field values that might contain RPM macros

gordonmessmer:
Assisted-by: Claude


hroncok (approved):
Looks OK. I haven't tested it, and the CI here is dead

gordonmessmer:
@gouldnicholas please review

